Issue #41, opened 2015-02-26
Grade School sorted-grade-school test.


sdroadie:
Currently, this test compares a sorted map against an unsorted hash map, where internal ordering isn't consistent. This needs to be changed.

The subsequent tests checks against a vector, which does maintain its internal order.

kytrinyx:
Good catch.

PR #42, opened 2015-02-26
Grade school sorted test now checks against a sorted map.

sdroadie:
I just opened issue #41 about this. It's a pretty small fix, but the test seems broken to me. Hash maps don't maintain their order, so the comparison fails. This should fix that.


kytrinyx:
Thanks!

PR #43, opened 2015-03-13
Add lein configuration files to exercises.

dkinzer:
By default lein sets source-path and test-path to "src" and "tests" 
respectively.  Therefore lein will require such folders to be set in order to
load any of the exercises such as are given in the clojure help section:

```
  To open a REPL using Leiningen change to the directory containing the
 exercise and run:

  $ lein repl

  Once you are ready to work on an exercise and have created a file to
 hold your solution (such as bob.clj) you can run the tests using
 load-file:

  (load-file "bob_test.clj")
```
But these instructions fail because lein doesn't know where to look for the
files.

This commit adds a lein config file for each of the exercises that forces lein
to look in the current directory for tests and source files.

This also has the added benefit that users that use vim-Fireplace will be able
to evaluate the code from within their editors.

### DIFF
--- a/accumulate/project.clj
+++ b/accumulate/project.clj
@@ -1,0 +1,6 @@
+(defproject accumulate "0.1.0-SNAPSHOT"
+  :description "accumulate exercise."
+  :url "https://github.com/exercism/xclojure/tree/master/accumulate"
+  :source-paths [""]
+  :test-paths [""]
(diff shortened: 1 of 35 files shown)


rubysolo:
Thank you!

dkinzer:
Cool, thanks!

PR #44, opened 2015-03-14
Remove lein config file from bin folder.

dkinzer:
The bin folder is not an exercise.  This commit removes a lein
configuration file that was incorrectly merged in by a previous pull
request.


kytrinyx:
Haha, yep :) Good call.

PR #48, opened 2015-04-07
Fixes pcalls, and cleans up tests with a macro.

alexkalderimis:
The main thing here is that the usage of `pcall` was incorrect. `pcall` takes a variable number of functions, not forms, and returns a lazy list, which must be evaluated to guarantee that all the calls are complete.

### DIFF
--- a/bank-account/bank_account_test.clj
+++ b/bank-account/bank_account_test.clj
@@ -1,5 +1,8 @@
 (ns bank-account-test
-  (:require [clojure.test :refer :all]))
+  (:refer-clojure :exclude (replace))
+  (:require [clojure.walk :refer (postwalk)]
+            [clojure.string :refer (capitalize replace)]
+            [clojure.test :refer :all]))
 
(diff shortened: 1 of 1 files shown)


kytrinyx:
@canweriotnow Would you mind reviewing this one?

canweriotnow:
@kytrinyx I will look at it tonight or tomorrow, my first reaction is that a macro is almost always the wrong thing, but I have to look at how these pcalls are working... my second reaction is that "almost" means the rest of the time, a macro is the _only_ thing :smile:

canweriotnow:
Actually, (and especially since it's in the test), I like the solution, and the fact is we don't exposed exercists to macros much at all, so it's good exposure but I don't think it affects the understandability of the test (at least not to my Clojurian eyes).

I would like it if @alexkalderimis would add comments or otherwise document what the macro is doing for the uninitiated, but I don't see that as blocking the merge.

PR #54, opened 2015-06-30
resolve warnings; use isLetter for readability

achengs:
The warnings were on Leiningen 2.5.1 on Java 1.8.0_45 OpenJDK 64-Bit Server VM:

call to static method isUpperCase on java.lang.Character can't be resolved (argument types: unknown).

call to static method isLowerCase on java.lang.Character can't be resolved (argument types: unknown).


rubysolo:
:+1: thank you.

PR #55, opened 2015-06-30
fewer conditionals --> clearer code

achengs:
Feel free to reject. Not sure what the purpose is of these examples but some of us currently feel this version is easier to grok than the other approach which uses many conditionals just to avoid data duplication.


kytrinyx:
At the moment the examples are intended as a sanity check for the test suite. At the moment they're not exposed.

PR #57, opened 2015-06-30
simpler code

achengs:


kytrinyx:
Thank you - this is much simpler, indeed :)

PR #58, opened 2015-06-30
simpler code

achengs:
no need to use seq, every?, set, let ... no need to repeat hash lookup


kytrinyx:
Thanks! In the future, would you mind putting the name of the exercise in the subject of the commit? It makes it a lot easier to find things when looking through the history.

My favorite post about git commit messages is this one by Tim Pope: http://tbaggery.com/2008/04/19/a-note-about-git-commit-messages.html

PR #73, opened 2015-10-14
Add strain exercise

cayennes:


cayennes:
I used the function names from the readme for this, which is kind of confusing in clojure since `keep` means something else.  But having different functions than the readme seemed even more confusing.  Was this the right choice?

Also I added an explicit test that the existing functions with the same behavior are not used, but didn't disallow other functions that would make it somewhat trivial (`for` with a `:when`.)

kytrinyx:
We could totally reword the README so it's more vague, describing only the behavior, and then saying each language will choose different terminology, in order to not crash with existing concepts.

I think it's fine to allow this to be a trivial exercise.

yurrriq:
Thanks, this looks good!

My only feedback would be to avoid "doesn't" in a symbol name and to use `(constantly ...)` instead of `(fn [& _] ...)`.
I feel reasonably strongly about the former, but the latter is perhaps a matter of personal preference.

If you could, make these changes or let me know here that you disagree with my suggestions and I'll merge.

```diff
diff --git a/strain/strain_test.clj b/strain/strain_test.clj
index 9b307b1..7500e91 100644
--- a/strain/strain_test.clj
+++ b/strain/strain_test.clj
@@ -32,10 +32,10 @@
 (deftest discard-several
   (is (= [0 2 4] (discard odd? (range 6)))))
 
-(deftest doesn't-use-existing-implementations
-  (with-redefs [filter (fn [& _] (throw (Exception. "Implement without filter!")))
-                remove (fn [& _] (throw (Exception. "Implement without remove!")))
-                filterv (fn [& _] (throw (Exception. "Implement without filterv!")))]
+(deftest does-not-use-existing-implementations
+  (with-redefs [filter  (constantly (throw (Exception. "Implement without filter!")))
+                remove  (constantly (throw (Exception. "Implement without remove!")))
+                filterv (constantly (throw (Exception. "Implement without filterv!")))]
     (dorun (retain even? (range 10)))
     (dorun (discard even? (range 10)))))
 ```

cayennes:
Done.  I have no idea why I didn't use `constantly` to begin with.

yurrriq:
I know the reason... Clojure eagerly evaluates `constantly`. That fact burns me all the time.. I tend to expect Haskell-like lazy behavior.

:+1: for `s/doesn't/does-not/` but you should revert the breaking usage of `constantly`. Sorry I led you astray...

Side note: It's awesome to have Travis tell us about regressions like this now :)

yurrriq:
I'm actually going to merge this then fix the mistake. Cheers!

PR #79, opened 2015-12-06
Use thrown-with-msg? in error case tests

awseward:
If the tests for error cases care only that a Throwable is thrown, there is potential for false passes to occur.

For example, if the implementation of `binary-search/search-for` throws an IndexOutOfBoundsException for some input, the test will pass, even though there's actually just a bug in `binary-search/search-for`.

By checking for some message, we can be a little more confident that the tests are passing only for the correct exceptions.

Also, updated example.clj to pass the modified tests again.


yurrriq:
Thanks!

Issue #49, opened 2015-04-10
largest_series_product is weird


alexkalderimis:
The expected values are very strange - `1` when no groups of a sufficient size can be produced, but `0` is a valid value. The strangeness of a sentinel value of `1` means that the obvious solution (below) does not pass the tests:

```clj
(defn largest-product [size input]
  (->> (map (zipmap "0123456789" (range)) input)
          (partition size 1)
          (map #(apply * %))
          (reduce max 0)))
```

Since you have to test if the input to reduce is empty, but cannot seed it with `1`. This extra (and pointless) complexity suggests that the task itself could do with improvement.

kytrinyx:
The identify of multiplication is 1: http://en.wikipedia.org/wiki/Identity_element
Multiplying by zero is still valid.

alexkalderimis:
It is not the zero that is the issue, but the fact that an empty set of runs is meant to produce a result of `1`. This doesn't have anything to do with multiplication, but the handling of an empty sequence.

kytrinyx:
I would think that this has everything to do with multiplication, since the problem is about multiplying numbers.

@exercism/track-maintainers Thoughts?

etrepum:
I agree with @kytrinyx, when the product of an empty list is defined to be a number, it should naturally be 1.

pminten:
Also agreed. This pattern/question comes up a lot actually, such as "what's the result of 'all elements of list X are true' if X is empty?"

In Haskell at least the rule seems to be "if it can be expressed as a repeated application of a binary function and has an identity element (i.e. it's a monoid) then the identity element is the result of passing in the empty list". This makes quite a lot of sense mathematically and I would advocate using this logic whenever the problem pops up in exercism.

alexkalderimis:
OK, I feel I'm going a bit mad here, so forgive the effort, but this really isn't a problem with multiplication, but with `max` .

The problem guarantees that we will never multiply empty lists (short groups are ignored), and the question isn't "how to I multiply an empty list?" (obviously return `1` - that is clear), but "what is the maximum of an empty list?". ie. there are two reductions here, and it is the second (the `max` one) that is annoying and counter-intuitive.

[...]

PR #85, opened 2015-12-16
Update largest-series-product per #49

yurrriq:
Clean up whitespace/line width and throw (and test for) an `Exception` when the given digit series is empty.


yurrriq:
Will someone else give this a once-over and make any changes they see fit?

PR #89, opened 2015-12-29
Add trinary exercise

haus:
This adds a trinary exercise to the clojure track. The tests were ported
from [xruby](https://github.com/exercism/xruby/blob/master/trinary/trinary_test.rb).

### DIFF
--- a/config.json
+++ b/config.json
@@ -26,6 +26,7 @@
     "scrabble-score",
     "roman-numerals",
     "binary",
+    "trinary",
     "prime-factors",
     "raindrops",
     "allergies",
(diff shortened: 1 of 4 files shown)


yurrriq:
It might be cool to do this in one pass, rather than `re-seq`, `num->digits`, `reverse` `map-indexed` and `apply`. [`.compareTo`][1] can come in handy too.

[1]: http://docs.oracle.com/javase/8/docs/api/java/lang/Character.html#compareTo-java.lang.Character-

yurrriq:
I went ahead and made the changes and merged #90. Thanks!

haus:
Ah nice. compareTo avoids the need turn it into an integer. I like the destructuring too; it didn't occur to me to use it with strings.

yurrriq:
Gotta love the `seq` abstraction.

PR #105, opened 2016-03-13
largest-series-product: Do not test digits/slices

petertseng:
The digits and slices functions are internal implementation details and
thus the test case for the largest-series-product problem should not be
concerned with testing them. The series tests could potentially be moved
to the series exercise (which would have to be a new addition to this track)

Their presence may cause students to falsely think that their solution
has to use these two functions, instead of the alternative
implementation of only iterating through the digits once.

If it is desired to give hints on how to approach this problem (which
was one advantage of having the digits and slices tests), then consider
including a hints file and/or directory in the largest-series-product
directory.

This PR arises from discussion in
https://github.com/exercism/x-common/issues/192

### DIFF
--- a/exercises/largest-series-product/src/example.clj
+++ b/exercises/largest-series-product/src/example.clj
@@ -1,8 +1,8 @@
 (ns largest-series-product)
 
-(defn digits [ds] (map #(Character/digit % 10) ds))
+(defn- digits [ds] (map #(Character/digit % 10) ds))
 
-(defn slices [n ds] (partition n 1 (digits ds)))
+(defn- slices [n ds] (partition n 1 (digits ds)))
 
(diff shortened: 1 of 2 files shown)


yurrriq:
What is the series exercise? I'm not seeing it in the track. This PR looks good though, thanks.

petertseng:
Oops. I had updated the commit text but forgot to update the PR text. The series problem would have to be a new addition to this track.

yurrriq:
Ah, cool. Feel free to make a PR to add series to xclojure if you're so inclined.

petertseng:
I will put it in my todo list. There are various things there so I may not get to that for an extended time. If anyone reading this issue cares, I would not mind if I were beaten to the punch on it.

PR #113, opened 2016-04-24
Remove check for sorted input from binary-search

### LINKED ISSUE #112: Binary search: check for non-sorted input defeats the purpose of binary search
Checking for non-sorted input is an O(n) operation. Performing the actual binary search is O(log(n)). Forcing the inclusion of an O(n) operation in the implementation of an O(log(n)) algorithm is confusing.

If you don't know that your input is sorted, there's no point in doing a binary search (sorting or checking for sorted input takes at least as long as a linear search), so I think this requirement should be removed.

ryanplusplus:
Closes #112

### DIFF
--- a/exercises/binary-search/src/example.clj
+++ b/exercises/binary-search/src/example.clj
@@ -5,8 +5,6 @@
 
 (defn search-for
   [elem alist]
-  {:pre [(or (apply <= alist)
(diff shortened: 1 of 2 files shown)


yurrriq:
LGTM, thanks!

Issue #114, opened 2016-04-26
Fix robot-name test suite false positive when using atoms.


dkinzer:
The following code passes current tests for the robot-name exercise on my computer:

```clojure
(ns robot-name)

(defn robot []
  (atom {:name "KIN123"}))

(defn robot-name [robot]
  (@robot :name))

(defn reset-name [robot]
  (swap! robot (fn [m] (assoc m :name "KIN125"))))
```

yurrriq:
The `robot-name-test` is flawed because `(robot-name/reset-name robbie)` should not be evaluated at the top level.

yurrriq:
Thank you raising this issue. I didn't mean to be overly terse. I'm at work. :smile:

PR #115, opened 2016-04-26
Fix false positive when using atoms.

dkinzer:
Ref #114

Use the lexical scope to prevent false positive.


yurrriq:
This dodges the real issue. `(robot-name/reset-name robbie)` should be encapsulated in a test (group).

yurrriq:
Thanks again for bring this problem to light.

PR #122, opened 2016-07-28
Add a stub exercise file to hello_world

tejasbubane:
This will make sure newcomers know where the code goes.

I am not sure if this is how we should do it. But I see other languages doing the same [eg. javascript](https://github.com/exercism/xjavascript/blob/master/exercises/bob/bob.js). Let me know if this will work and I will add stubs for other exercises. Not sure if we should do this for all exercises or few initial ones. I don't see ruby/javascript doing it for all. 

Also we should have something like this for `hello-world` https://github.com/exercism/xjavascript/blob/master/exercises/hello-world/HINTS.md

### DIFF
--- a/exercises/hello-world/HINTS.md
+++ b/exercises/hello-world/HINTS.md
@@ -1,0 +1,28 @@
+### Project Structure
+
+Clojure exercises in exercism use [leiningen](http://leiningen.org/) to configure and run your code
+and use [leiningen standard directory structure](https://github.com/technomancy/leiningen/blob/master/doc/TUTORIAL.md#directory-layout).
+
+You will find a test file named `hello_world_test.clj` inside `test` directory.
+Write your code in `src/hello_world.clj`. It should use the namespace `hello-world` so that tests can pick it up.
+
+### Running tests
+
+Run the tests using `lein test` command and make them pass:
+
+```
+$ lein test
+
+lein test hello-world-test
+
+Ran 3 tests containing 3 assertions.
+0 failures, 0 errors.
+```
+
+Then submit the exercise using:
+
+```
+$ exercism submit src/hello_world.clj
+```
+
+For more detailed instructions and learning resources refer [exercism's clojure language page](http://exercism.io/languages/clojure).
--- a/exercises/hello-world/src/hello_world.clj
+++ b/exercises/hello-world/src/hello_world.clj
@@ -1,0 +1,5 @@
+(ns hello-world)
+
+(defn hello [] ;; <- arglist goes here
+  ;; your code goes here
+)


yurrriq:
Thanks for working on this!

My only suggestion would be to consider adding something like ` ;; arglist goes here` where appropriate. I could see newcomers getting potentially misled as it is. Maybe that's not a real concern...

tejasbubane:
@yurrriq Added `arglist` to the stub file and a `getting_started.md` file. I guess this file will get appended to the `readme` when user will fetch the exercise.

tejasbubane:
Should we add `stub` files to a few other exercises after `hello-world` as well?

yurrriq:
I don't think `GETTING_STARTED.md` belongs in each exercise directory, but the arglist addition looks good to me.

Any opinions, @kytrinyx @exercism/lisp?

tejasbubane:
> I don't think GETTING_STARTED.md belongs in each exercise directory

By `stub` files, I meant solution file inside `src` directory for a few more exercises after `hello_world`. `GETTING_STARTED.md` should be for first exercise only.

yurrriq:
Aha. I think traditionally that sort of file lives in the docs directory at the top level of this repo, but I can see the value of including it with the first exercise too.

tejasbubane:
[Ruby](https://github.com/exercism/xruby/blob/master/exercises/hello-world/GETTING_STARTED.md) and [Javascript](https://github.com/exercism/xjavascript/blob/master/exercises/hello-world/HINTS.md) have them (of the ones I know). 
Also I just verified on javascript track that the file gets appended to `README.md` when a user fetches the exercise.

verdammelt:
Putting a 'hints' or 'getting started' into the hello world exercise could be useful - but I'd disagree with putting it in *each* exercise.

tejasbubane:
> but I'd disagree with putting it in each exercise.

Definitely not each exercise, just the first one - `hello-world`.

kytrinyx:
If we rename `GETTING_STARTED.md` to `HINTS.md` the contents of it will be appended to the README (just for this exercise). We didn't have that when ruby, javascript, etc made a getting started file.

kytrinyx:
> I guess this file will get appended to the readme when user will fetch the exercise.

It will if you rename it to `HINTS.md` (that is not documented very well).

yurrriq:
Cool. Thanks for the feedback. Sounds like if you rename it to`HINTS.md` this'll be good to go. :+1: the stubs for each exercise if you want to take that on too, @tejasbubane. Thanks for all your help!

tejasbubane:
> If we rename GETTING_STARTED.md to HINTS.md the contents of it will be appended to the README

I assumed any markdown file(s) in the directory would get appended. I felt I had read this somewhere in the comments long back, but seems like I was wrong. Nevermind.

@yurrriq Renamed to `HINTS.md`. Will add some more stubs in another PR. Keeping this one for `hello-world` only.

yurrriq:
Sounds good! Thanks again.

PR #127, opened 2016-08-23
Add bracket-push

leetwinski:
added bracket-push exercise

### DIFF
--- a/exercises/bracket-push/src/example.clj
+++ b/exercises/bracket-push/src/example.clj
@@ -1,0 +1,11 @@
+(ns bracket-push)
+
+(defn valid? [s]
+  (let [pairs { \) \( \] \[ \} \{ }]
+    (loop [stack [] [x & xs :as s] s]
+      (cond (empty? s) (empty? stack)
+            (#{ \( \[ \{ } x) (recur (conj stack x) xs)
+            (#{ \) \] \} } x) (if (= (peek stack) (pairs x))
+                                (recur (pop stack) xs)
+                                false)
+            :else (recur stack xs)))))
--- a/exercises/bracket-push/src/example.clj
+++ b/exercises/bracket-push/src/example.clj
@@ -1,11 +1,13 @@
 (ns bracket-push)
 
 (defn valid? [s]
-  (let [pairs { \) \( \] \[ \} \{ }]
+  (let [pairs { \) \( \] \[ \} \{ }
+        opening (set (vals pairs))
+        closing (set (keys pairs))]
     (loop [stack [] [x & xs :as s] s]
       (cond (empty? s) (empty? stack)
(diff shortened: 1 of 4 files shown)


kytrinyx:
Thank you @leeqleeq, @leetwinski!

@exercism/clojure if you have a moment to review the code, that would be great.

exercises/bracket-push/src/example.clj
```diff
@@ -0,0 +1,11 @@
+(ns bracket-push)
+
+(defn valid? [s]
+  (let [pairs { \) \( \] \[ \} \{ }]
+    (loop [stack [] [x & xs :as s] s]
+      (cond (empty? s) (empty? stack)
+            (#{ \( \[ \{ } x) (recur (conj stack x) xs)
+            (#{ \) \] \} } x) (if (= (peek stack) (pairs x))
```
yurrriq: Is there any way to avoid the explicit `false` here? Perhaps you could use `when` and then call `(bool (recur stack xs))`. I'm not sure I like that any better.

exercises/bracket-push/src/example.clj
```diff
@@ -0,0 +1,11 @@
+(ns bracket-push)
+
+(defn valid? [s]
+  (let [pairs { \) \( \] \[ \} \{ }]
+    (loop [stack [] [x & xs :as s] s]
+      (cond (empty? s) (empty? stack)
+            (#{ \( \[ \{ } x) (recur (conj stack x) xs)
```
yurrriq: Is there a good way to avoid manually repeating the bracket characters here? Perhaps `pairs` could be a map and the `cond` clauses could use `(keys pairs)` and `(values pairs)`. Again, I'm not sure if I prefer that.

yurrriq:
I've added a few comments/questions inline, but other than that this looks great. Thanks.

exercises/bracket-push/src/example.clj
leetwinski: as far as i remember, `(bool (recur ...))` wont work, since it is not tail recursive. As for `when`, i would probably prefer this variant, limiting return value no bollean, not boolean'ish.

exercises/bracket-push/src/example.clj
leetwinski: right. i already have `pairs` there. Will reuse that

exercises/bracket-push/src/example.clj
yurrriq: Quite right. Why not put the `bool` call just outside `loop` then?

leetwinski:
@yurrriq , added sets for `opening`/`closing` brackets.

yurrriq:
Thanks. LGTM. I think `false` is maybe more clear than `bool` anyway. Merging.

PR #131, opened 2016-08-26
[pangram] exercise

leetwinski:


leetwinski:
@exercism/clojure , added pangram exercise

exercises/pangram/src/example.clj
```diff
@@ -0,0 +1,15 @@
...
+(def ^:private z-int (int \z))
+
+(def ^:private alphabet (set (range a-int (inc z-int))))
+
+(defn pangram? [input]
+  (= alphabet
+     (->> input
+          lower-case
+          (map int)
+          (filter #(<= a-int % z-int))
```
yurrriq: Why not [`Character/isLowerCase`](http://docs.oracle.com/javase/8/docs/api/java/lang/Character.html#isLowerCase-char-) or regex here instead?

exercises/pangram/src/example.clj
```diff
@@ -0,0 +1,15 @@
+(ns pangram
+  (:require [clojure.string :refer [lower-case]]))
+
+(def ^:private a-int (int \a))
+(def ^:private z-int (int \z))
+
+(def ^:private alphabet (set (range a-int (inc z-int))))
+
+(defn pangram? [input]
+  (= alphabet
```
yurrriq: Why not fold the `(= alphabet)` into the `(->> ...)`?

exercises/pangram/src/example.clj
```diff
@@ -0,0 +1,15 @@
+(ns pangram
+  (:require [clojure.string :refer [lower-case]]))
+
+(def ^:private a-int (int \a))
+(def ^:private z-int (int \z))
+
+(def ^:private alphabet (set (range a-int (inc z-int))))
```
yurrriq: `a-int`, `z-int` and `alphabet` feel superfluous to me.

exercises/pangram/src/example.clj
yurrriq: ... or something like:

```clojure
(def char<=? (comp (complement pos?) compare))

(char<=? \a \z)
;; => true
(char<=? \a \a)
;; => true
(char<=? \z \a)
;; => false
```

exercises/pangram/test/pangram_test.clj
```diff
@@ -0,0 +1,37 @@
...
+  (is
+   (false?
+    (pangram? "7h3 qu1ck brown fox jumps ov3r 7h3 lazy dog"))))
+
+(deftest mixed-case-and-punctuation
+  (is (pangram? "\"Five quacking Zephyrs jolt my wax bed.\"")))
+
+(deftest non-ascii-characters
+  (is (pangram?
+       "Victor jagt zwölf Boxkämpfer quer über den großen Sylter Deich.")))
```
yurrriq: Given `alphabet` and `#(<= a-int % z-int)` above, is the intended behaviour to ignore accented letters?

yurrriq:
What are your thoughts on something like this?

```clojure
(ns pangram
  (:require [clojure.string :refer [lower-case]]
            [clojure.set    :refer [subset?]]))

(def ^:private alphabet (set "abcdefghijklmnopqrstuvwxyz"))

(defn pangram? [input]
  (-> (fn [acc chr]
        (cond-> acc
          (Character/isLetter chr) (conj (Character/toLowerCase chr))))
      (reduce #{} input)
      (->> (subset? alphabet))))
```

This way we only traverse `input` once and use `#'clojure.set/subset?`, rather than requiring equality.

exercises/pangram/test/pangram_test.clj
leetwinski: Yes, it should just consider the letters from the english alphabet. The rest is considered to be junk

exercises/pangram/src/example.clj
leetwinski: it should be. Thanks!

leetwinski:
I don't know. As for me, i always find it `cond->` much harder to read. I would probably rewrite it to use `(= count 26)` as the correct solution indicator

yurrriq:
I think folding over a set is the best solution, but have no strong feelings re: `cond->`. It could be a good use case for transducers too and it might be good to avoid the Java methods. `compare` works in cljs too fwiw.

leetwinski:
@yurrriq , yet another variant: transducer + characters count.

leetwinski:
Good point about clojurescript. Would rewrite `in-range?`

yurrriq:
Nice. I like `char<=`. I still think we should be using sets and `#'clojure.set/subset?` rather than collecting a vector and checking its size.

Thoughts, @exercism/lisp?

leetwinski:
I just feel uncomfortable about the double traversal (reduction + subset?) vs single pass with transducer and count. But I won't really insist if you say your variant is better, since for the values under test the performance doesn't matter i guess. Anyway, i can always post my answer as the solution straight to exercism :smile:

yurrriq:
Fair point. `count` for vectors is O(1), right?

leetwinski:
yes, since it is `counted`

yurrriq:
My only other suggestion then would be to try to avoid `Character/toLowerCase` for portability. I'm happy to merge this now. What do you think?

leetwinski:
The only thing i can think of, is to use `clojure.string/lower-case` on a char, but it seems to be an overhead. Maybe we should go with conditional compilation and rename example to `.cljc` ?

yurrriq:
If you want to. Otherwise I think it's fine to leave it as it is, since we don't really mention portability anywhere. If anyone asks we can point to this discussion.

leetwinski:
Ok then. I wouldn't insist on rewriting :smile:

yurrriq:
Thanks for this!

PR #137, opened 2016-09-26
Add all-your-base exercise

haus:
This commit adds the all-your-base exercise, with the tests ported from
the ruby version of the exercise.

### DIFF
--- a/config.json
+++ b/config.json
@@ -45,6 +45,7 @@
     "binary-search-tree",
     "difference-of-squares",
     "hexadecimal",
+    "all-your-base",
     "largest-series-product",
     "pov",
     "bracket-push",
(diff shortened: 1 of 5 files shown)


yurrriq:
Hey, thanks! First, I think you'll need to rename `all_your_base.clj` to `example.clj` to get the tests passing. And then I think it'd be good to keep the function arguments on one line instead of one per argument, e.g.

```clj
[x y z]
;; instead of
[x
 y
 z]
```

yurrriq:
The `(->> ...)` form in the `:else` clause of `convert`'s `cond` looks misaligned, at least on mobile, so maybe double check that too and reindent as necessary.

haus:
Thanks for the quick feedback. I pushed out a commit that should resolve it. Happy to squash that into the previous commit if travis passes.

yurrriq:
LGTM, thanks!

PR #138, opened 2016-09-26
Add flatten-array exercise

haus:
This adds the flatten-array exercise, with tests based on the canonical
test data.

### DIFF
--- a/config.json
+++ b/config.json
@@ -83,6 +83,11 @@
     },
     {
       "difficulty": 1,
+      "slug": "flatten-array",
+      "topics": []
+    },
+    {
+      "difficulty": 1,
       "slug": "binary-search",
       "topics": []
     },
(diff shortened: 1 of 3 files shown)


yurrriq:
Should we try to implement this more efficiently, e.g. Using `recur`, lazy sequences and/or `(filter some? ...)`?

haus:
That sounds good to me. I'll take a stab at it using recur and filter.

haus:
Okay, here is another attempt, this time with recur and remove used. I'm not sure how much better it is though.

yurrriq:
Thanks. I should be able to take a proper look in the next few days of no one beats me to it.

yurrriq:
Have you made a final decision here?

haus:
I'm reasonably happy with the current version (loop/recur), unless you had any concerns.

I can rebase against the updated config.json in any case.

mwfogleman:
Hm - I think the original solution is more readable, even if it's less efficient. However, it is possible to have a solution that is both readable and efficient, using tree-seq, which is how the core flatten is implemented. I will make a PR that does that.

(FYI, a blog post that I wrote a couple of years ago talks about a few ways to implement flatten. https://mwfogleman.github.io/posts/20-12-2014-flatcat.html)

PR #140, opened 2016-10-04
implemented exercise isogram

querenker:


exercises/isogram/src/isogram.clj
```diff
@@ -0,0 +1,5 @@
+(ns isogram
```
yurrriq: This file needs to be renamed to `example.clj`.

config.json
```diff
@@ -52,7 +52,8 @@
     "minesweeper",
     "change",
     "pangram",
-    "wordy"
+    "wordy",
+	"isogram"
```
yurrriq: The indentation of `"isogram"` needs to be adjusted to match that of the rest of the `"problems"`.

querenker: @yurrriq Thank you for your annotations. Now it should work and look properly :)

yurrriq:
Thanks! LGTM.

PR #141, opened 2016-10-08
Re-add bank account

haus:
This PR re-adds the previous bank-account exercise with an updated
test suite. The previous test suite had a call to (shutdown-agents)
outside of a function body, which caused it to get called before all
tests had executed. This version of the test suite uses a fixture to
ensure that the (shutdown-agents) call happens *after* all tests have
executed.

### DIFF
--- a/config.json
+++ b/config.json
@@ -33,6 +33,7 @@
     "octal",
     "strain",
     "atbash-cipher",
+    "bank-account",
     "crypto-square",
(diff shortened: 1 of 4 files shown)


yurrriq:
Thanks for doing this! Looks like Travis is pleased too :smile:

yurrriq:
Thanks!

PR #147, opened 2016-10-22
implemented exercise acronym

querenker:


yurrriq:
Thanks, I'll try to review this weekend.

exercises/acronym/src/example.clj
```diff
@@ -0,0 +1,8 @@
+(ns acronym
+  (:require [clojure.string :as str]))
+
+(defn acronym [text]
+  (->> (re-seq #"[A-Z]+[a-z]*|[a-z]+" text)
+       (map first)
+       (map str/upper-case)
```
yurrriq: Why not do `(->> ... (apply str) (str/upper-case))`?

exercises/acronym/src/example.clj
```diff
@@ -0,0 +1,8 @@
+(ns acronym
+  (:require [clojure.string :as str]))
+
+(defn acronym [text]
+  (->> (re-seq #"[A-Z]+[a-z]*|[a-z]+" text)
+       (map first)
```
yurrriq: Couldn't we use group capturing to obviate the need for `(->> ... (map first) ...)`?

exercises/acronym/src/example.clj
querenker: Thanks, that's a good point and I'll fix this

exercises/acronym/src/example.clj
querenker: I can't find a good solutions without impairing the readability. But maybe I miss something (I am relatively new to Clojure)

yurrriq:
:+1: for favoring readability over micro-optimizations. I think this is good to merge then if you do.

haus:
@querenker looks like the PR has a conflict with config.json. If you're able to resolve it, I can merge this in. If you need a hand resolving the conflict, let me know.

querenker:
Now the merge conflict should be solved.

haus:
Thanks @querenker !

PR #151, opened 2016-11-03
Simplify example implementation of robot-name

nwallace:
The original implementation seemed overly-complicated to me. This changes the example so that you generate the name when you generate the robot, rather than when you ask for its name.

### DIFF
--- a/exercises/robot-name/src/example.clj
+++ b/exercises/robot-name/src/example.clj
@@ -7,10 +7,10 @@
        (+ 100 (.nextInt random 899))))
 
 (defn robot []
-  {:name (atom (generate-name))})
+  (atom {:name (generate-name)}))
 
 (defn robot-name [robot]
-  @(:name robot))
+  (:name @robot))
 
 (defn reset-name [robot]
-  (reset! (:name robot) (generate-name)))
+  (swap! robot assoc :name (generate-name)))
--- a/exercises/robot-name/src/example.clj
+++ b/exercises/robot-name/src/example.clj
@@ -1,19 +1,16 @@
 (ns robot-name)
 
-(defn robot []
-  {:name (atom "")})
-
-(def random (java.util.Random.))
-(def letters (map char (range 65 91)))
-(defn generate-name []
+(def ^:private random (java.util.Random.))
+(def ^:private letters (map char (range 65 91)))
+(defn- generate-name []
   (str (apply str (take 2 (shuffle letters)))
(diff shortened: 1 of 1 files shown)


yurrriq:
Should the whole robot be an atom instead of just the name? I think @canweriotnow commented about this somewhere.

nwallace:
@yurrriq Ah yes, I read that the wrong way round. And yeah that makes sense. Just pushed an update

exercises/robot-name/src/example.clj
```diff
@@ -1,19 +1,16 @@
...
-  {:name (atom "")})
-
-(def random (java.util.Random.))
-(def letters (map char (range 65 91)))
-(defn generate-name []
+(def ^:private random (java.util.Random.))
+(def ^:private letters (map char (range 65 91)))
+(defn- generate-name []
   (str (apply str (take 2 (shuffle letters)))
        (+ 100 (.nextInt random 899))))
```
haus: So I *think* the addition here is to ensure that the number is 3 digits, but the result is that 000 through 099 won't be present in robot names. Also, the argument for nextInt is exclusive, so 899 will never be selected (https://docs.oracle.com/javase/7/docs/api/java/util/Random.html).

Something like `(format "%03d" (.nextInt random 1000))` should work and use the full range of 000 to 999.

exercises/robot-name/src/example.clj
```diff
@@ -1,19 +1,16 @@
...
-(defn robot []
-  {:name (atom "")})
-
-(def random (java.util.Random.))
-(def letters (map char (range 65 91)))
-(defn generate-name []
+(def ^:private random (java.util.Random.))
+(def ^:private letters (map char (range 65 91)))
+(defn- generate-name []
   (str (apply str (take 2 (shuffle letters)))
```
haus: This `(take 2 (shuffle letters)))` call never returns a prefix of the AA, BB, CC... type. In order to include all combinations of characters, something like `(apply str (repeatedly 2 #(first (shuffle letters))))` should work.

haus:
I think the changes you made here look really good btw.

haus:
I realize the comments I left were on lines you didn't change, so if you don't get the chance I'd be happy to make a PR to update those once this is merged.

haus:
Thanks for this @nwallace !

haus:
I opened a PR with the unrelated changes I suggested here: https://github.com/exercism/xclojure/pull/157.

PR #152, opened 2016-11-11
Remove unsorted-vector variable

kevgathuku:
Seems to be leftover from #113


kytrinyx:
Nice catch, thanks!

kevgathuku:
You're welcome ☺

PR #153, opened 2016-11-12
Added pig-latin exercise

lmccombes:
I've created some tests using the sample data at https://github.com/exercism/x-common/blob/master/exercises/pig-latin/canonical-data.json and have added an example solution.

### DIFF
--- a/exercises/pig-latin/src/example.clj
+++ b/exercises/pig-latin/src/example.clj
@@ -1,45 +1,47 @@
 (ns pig-latin
   (:require [clojure.string :as str]))
 
-(defn starts-with-any [prefixes]
-  (fn [word]
-    (some (partial str/starts-with? word) prefixes)))
+(defn starts-with-any [prefixes word]
+  (some (partial str/starts-with? word) prefixes))
 
 (defn- starts-with-vowel-like? [word]
-  ((starts-with-any #{"yt" "xr"}) word))
+  (starts-with-any #{"yt" "xr"} word))
 
 (defn starts-with-vowel? [word]
-  ((starts-with-any #{"a" "e" "i" "o" "u"}) word))
+  (starts-with-any #{"a" "e" "i" "o" "u"} word))
 
 (defn- starts-with-two-letter-prefix? [word]
-  ((starts-with-any #{"ch" "qu" "th"}) word))
+  (starts-with-any #{"ch" "qu" "th"} word))
 
 (defn- starts-with-three-letter-prefix? [word]
-  ((starts-with-any #{"thr" "sch"}) word))
+  (starts-with-any #{"thr" "sch"} word))
 
 (defn- starts-with-qu-and-preceeding-constanant? [word]
   (and (not (starts-with-vowel? word))
        (str/starts-with? (subs word 1) "qu")))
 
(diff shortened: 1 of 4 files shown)


kytrinyx:
Thanks so much, @lmccombes! @haus would you have a moment to take a look at this?

haus:
@kytrinyx surely

exercises/pig-latin/src/example.clj
```diff
@@ -0,0 +1,47 @@
...
+  ((starts-with-any #{"ch" "qu" "th"}) word))
+
+(defn- starts-with-three-letter-prefix? [word]
+  ((starts-with-any #{"thr" "sch"}) word))
+
+(defn- starts-with-qu-and-preceeding-constanant? [word]
+  (and (not (starts-with-vowel? word))
+       (str/starts-with? (subs word 1) "qu")))
+
+(defn rotate [word n]
```
haus: It looks like rotate can be a private function.

exercises/pig-latin/src/example.clj
```diff
@@ -0,0 +1,47 @@
+(ns pig-latin
+  (:require [clojure.string :as str]))
+
+(defn starts-with-any [prefixes]
```
haus: I think this would be more readable as a two argument function instead of a single arg function returning another single arg function. It could be something like `starts-with-any? [prefixes word]`.

haus:
@lmccombes Thanks so much! I left a couple of comments, and then I think this should be good to go.

lmccombes:
I think the edits you suggested make sense, I've added them in a new commit.

exercises/pig-latin/src/example.clj
```diff
@@ -0,0 +1,49 @@
...
+  (starts-with-any #{"thr" "sch"} word))
+
+(defn- starts-with-qu-and-preceeding-constanant? [word]
+  (and (not (starts-with-vowel? word))
+       (str/starts-with? (subs word 1) "qu")))
+
+(defn- rotate [word n]
+  (str (subs word n) (subs word 0 n)))
+
+(defn append-ay [word]
```
haus: I think this can be private too right?

exercises/pig-latin/src/example.clj
```diff
@@ -0,0 +1,49 @@
+(ns pig-latin
+  (:require [clojure.string :as str]))
+
+(defn starts-with-any [prefixes word]
```
haus: One more potential private function.

exercises/pig-latin/src/example.clj
```diff
@@ -0,0 +1,49 @@
+(ns pig-latin
+  (:require [clojure.string :as str]))
+
+(defn- starts-with-any [prefixes word]
+  (some (partial str/starts-with? word) prefixes))
+
+(defn- starts-with-vowel-like? [word]
+  (starts-with-any #{"yt" "xr"} word))
+
+(defn starts-with-vowel? [word]
```
haus: Last one I think 😄

lmccombes: Thanks :)

haus:
Looks great! Thanks for the PR @lmccombes

PR #154, opened 2016-11-14
Add clojure test generator and clock exercise

haus:
This commit adds a generator namespace, adds a stencil dependency, and
adds x-common to the .gitignore. When invoked, this will clone the
x-common repo if not present, then load and munge any test-data for the
passed exercise, and then render clojure tests using a mustache template
for the exercise.

### DIFF
--- a/_src/generator.clj
+++ b/_src/generator.clj
@@ -1,0 +1,72 @@
+(ns generator
+  (:require [cheshire.core :as json]
+            [clojure.java.shell :refer [sh]]
+            [clojure.java.io :refer [reader]]
+            [stencil.core :as stencil])
+  (:import (java.io File IOException)))
+
+(defn file-exists?
+  "Helper function to determine if a given file exists."
+  [path]
+  (.exists (File. path)))
+
+(defn warn-and-exit
+  "Wrapper function to warn with the given message and then exit with a non-zero return"
+  [message]
+  (println message)
+  (System/exit 1))
+
+(defn clone-test-data
+  "Clone the x-common repo from github"
+  []
+  (sh "git" "clone" "git@github.com:exercism/x-common"))
+
+(defn load-test-data
+  "Clones and loads the test data for the given exercise"
+  [exercise-name]
+  (when-not (file-exists? "x-common")
+    (clone-test-data))
+  (let [test-data-filename (format "x-common/exercises/%s/canonical-data.json" exercise-name)]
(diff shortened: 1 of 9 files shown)


haus:
Happy to break this into 2 PRs if that makes more sense. I put them together to show the how the test generation would work.

_src/generator.clj
```diff
@@ -0,0 +1,56 @@
...
+
+(defn file-exists?
+  "Helper function to determine if a given file exists."
+  [path]
+  (.exists (File. path)))
+
+(defn clone-test-data
+  "Clone the x-common repo from github"
+  []
+  (sh "git" "clone" "git@github.com:exercism/x-common"))
```
haus: It looks like the ruby test generator doesn't actually clone the repo, but just warns and exits if the repo doesn't exist (https://github.com/exercism/xruby/blob/master/lib/generator.rb#L46). Would that be preferable here?

project.clj
```diff
@@ -2,5 +2,8 @@
   :description  "Exercism Exercises in Clojure"
   :url          "https://github.com/exercism/xclojure"
   :test-paths   ["_test"]
+  :source-paths ["_src"]
```
haus: I used `_src` here because `_test` had been used for tests previously. Happy to put it somewhere else as well.

haus:
Looks like travis may be sad because i added the _src directory.

yurrriq:
Looks like you need to add it to `config.json` maybe.

_src/generator.clj
```diff
@@ -0,0 +1,56 @@
...
+       (format "Could not find test data for %s (looking in %s)"
+               exercise-name test-data-filename)))
+    (json/parse-stream (reader test-data-filename))))
+
+(defn munge-test-data
+  "Loads the generator namespace for the exercise and calls the munge-data function on the given test-data."
+  [exercise-name test-data]
+  (let [exercise-ns (symbol (str exercise-name "-generator"))]
+    (require [exercise-ns])
+    (if-let [munge-data-fn (ns-resolve exercise-ns (symbol "munge-data"))]
```
haus: In the case of the clock exercise munging the canonical data wasn't necessary, but it seemed like a good escape hatch to leave in for other exercises, as mustache is purely data driven templating.

yurrriq:
I think this should be good to merge now. Go ahead if you agree.

haus:
Right on.

PR #157, opened 2016-11-24
Simplify generate-name function for robot-name

haus:
This commit updates the generate-name function in robot-name to use
clojure's rand-int function instead of java interop. It also updates the
rand-int call to include all of the range 0-999 and refactors the string
assembly somewhat.


yurrriq:
How about `(rand-nth letters)` instead of `(first (shuffle letters))`?

haus:
good call

haus:
@yurrriq any reason I shouldn't merge this in?

yurrriq:
I don't think so

PR #163, opened 2017-04-14
secret handshake exercise

shdblowers:
Adding the secret handshake exercise.

This is my first exercism exercise submission, not sure if I have everything right.

Also, still new to Clojure so the example isn't great :smile:

### DIFF
--- a/exercises/secret-handshake/src/example.clj
+++ b/exercises/secret-handshake/src/example.clj
@@ -6,9 +6,9 @@
       (reverse)))
 
 (defn- convert [integer]
-  (filter some?
+  (remove nil?
           (map (fn [command binary]
-                 (if (= \1 binary) command))
+                 (when (= \1 binary) command))
                ["wink" "double blink" "close your eyes" "jump"]
                (int->reversed-binary integer))))
 
--- a/exercises/secret-handshake/src/example.clj
+++ b/exercises/secret-handshake/src/example.clj
@@ -1,17 +1,19 @@
 (ns secret-handshake)
 
-(defn- to-binary-seq [integer]
-  (reverse (seq (Integer/toString integer 2))))
+(defn- int->reversed-binary [int]
+  (-> int
+      (Integer/toBinaryString)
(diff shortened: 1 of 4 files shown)


shdblowers:
@haus what do you think of this PR?

haus:
@shdblowers thanks for the contribution! I'm not sure how I missed seeing it pop up, but I'll take a look at it over this weekend if that works for you.

shdblowers:
Accidentally closed it.

Thanks @haus over the weekend is perfect.

shdblowers:
Thanks for the comments @haus, agree with all of them, have updated the PR to reflect them.

exercises/secret-handshake/src/example.clj
```diff
@@ -0,0 +1,19 @@
...
+
+(defn- int->reversed-binary [int]
+  (-> int
+      (Integer/toBinaryString)
+      (reverse)))
+
+(defn- convert [integer]
+  (filter some?
+          (map (fn [command binary]
+                 (if (= \1 binary) command))
```
haus: Since there is no `else` for this `if`, you can just use `when`.

exercises/secret-handshake/src/example.clj
```diff
@@ -0,0 +1,19 @@
+(ns secret-handshake)
+
+(defn- int->reversed-binary [int]
+  (-> int
+      (Integer/toBinaryString)
+      (reverse)))
+
+(defn- convert [integer]
+  (filter some?
```
haus: this could be `remove nil?` instead of `filter some?`. I'm not sure which is more readable.

haus:
@shdblowers it's looking real good. I left comments for 2 other things i noticed. I think this is ready to merge either way.

shdblowers:
I think `remove nil?` sounds more readable, so have changed it to that.

Also did the change of `if` to `when`.

shdblowers:
I'm good. Let's merge this bad boy! 😄

PR #165, opened 2017-05-06
Rename docs files to .md

### LINKED ISSUE #158: Verify contents and format of track documentation
Each language track has documentation in the `docs/` directory, which gets included on the site
on each track-specific set of pages under [/languages](http://exercism.io/languages).

We've added some general guidelines about how we'd like the track to be documented in exercism/exercism.io#3315
which can be found at https://github.com/exercism/exercism.io/blob/master/docs/writing-track-documentation.md

Please take a moment to look through the documentation about documentation, and make sure that
the track is following these guidelines. Pay particularly close attention to **how to use images**
in the markdown files.

Lastly, if you find that the guidelines are confusing or missing important details, then a pull request
would be greatly appreciated.

haus:
Previously they had a .org extension, which hides their format of
markdown, so this commit moves them to the md extension.

Should resolve #158 .


yurrriq:
Their format is [Org](http://orgmode.org), not Markdown.

yurrriq:
I'm closing this. If you have a compelling case to switch back to Markdown, you should open a separate PR and actually convert, rather than just rename.

haus:
Oh good call, I didn't realize org was a format. thanks!

PR #170, opened 2017-06-12
Garden fixes

mwfogleman:
The main fix I wanted to fix is to use clojure.string/lower-case rather than the Java method, as that is more idiomatic in this case. But I also did some formatting fixes.


yurrriq:
This looks good, thanks. You'll need to rebase before I can merge, but I honk this also might be a good opportunity to rewrite a few forms with `->>` for increased legibility.

mwfogleman:
Great! Threaded and rebased. :)

yurrriq:
Thanks!

PR #171, opened 2017-06-16
Use tree-seq in flatten-array.

mwfogleman:
Use tree-seq for readability and efficiency, re: my comment in #138.


yurrriq:
LGTM /cc @haus

haus:
Ditto. Thanks @mwfogleman !

PR #173, opened 2017-06-18
Add rotational-cipher puzzle with tests and solution.

mwfogleman:
Added the rotational-cipher puzzle with tests and solution.

https://github.com/exercism/x-common/blob/master/exercises/rotational-cipher/description.md

### DIFF
--- a/config.json
+++ b/config.json
@@ -422,6 +422,13 @@
       "topics": [
 
       ]
+    },
+    {
+      "difficulty": 1,
+      "slug": "rotational-cipher",
+      "topics": [
+
+      ]
     }
   ]
 }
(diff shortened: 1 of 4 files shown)


yurrriq:
Thanks. This looks pretty good, but could you please make the internal definitions private?

It would also be good to put blank lines between the top-level definitions, but I don't feel strongly about that, so it's up to you.

I also wonder if we could make use of `mapcat`.

mwfogleman:
I added blank lines, made the internal function private (but not the variables), and added the metadata to the config.json file.

The solution does not include mapcat at this time.

yurrriq:
Looks like `config.json` is missing a `,` and causing the Travis build to fail. Could you please add `^:private` to the `def`s too? I'm fine with the current `mapcat`-less implementation.

mwfogleman:
Ah, that's why! Thanks. Done and done. :)

yurrriq:
LGTM, thanks!

PR #174, opened 2017-07-02
Test the cipher for opposite direction rotation.

dkinzer:
A true rotational cipher does not care whether one rotates forwards or backwards.  The solution should be tested for this case.

### DIFF
--- a/exercises/rotational-cipher/src/example.clj
+++ b/exercises/rotational-cipher/src/example.clj
@@ -1,21 +1,27 @@
 (ns rotational-cipher)
 
-(def ^:private string (partial apply str))
+(defn ^:private spinner [s a]
+  (let [a (int a)
+        spin (fn [c] (mod (+ c s) 26))]
+    (fn [c]
+      (let [c (- (int c) a)]
+        (char (+ (spin c) a))))))
 
-(def ^:private lower-case "abcdefghijklmnopqrstuvwxyz")
+(defn ^:private upper-spinner [s]
+  (spinner s \A))
 
-(def ^:private upper-case (->> lower-case (map clojure.string/upper-case) string))
+(defn ^:private lower-spinner [s]
+  (spinner s \a))
 
-(def ^:private letters (into #{} (concat lower-case upper-case)))
+(defn ^:private cipher [spin]
+  (let [upper-spin (upper-spinner spin)
+        lower-spin (lower-spinner spin)]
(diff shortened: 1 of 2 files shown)


yurrriq:
Thanks. At first glance, this looks good to me. Is there any inter-track precedence for this? If so, are we following it?

dkinzer:
To me it looks like the only other track that has this exercise so far is the #C track with 105 problems.  F# with 104 exercises still doesn't have it.  Also Ruby and Go don't have the exercise.

The C# exercise does not currently test for backwards rotation either (fwiw).  I don't know C#, or I would probably post a PR there too.

yurrriq:
Hmm... to me [rotational-cipher](https://github.com/exercism/problem-specifications/blob/master/exercises/rotational-cipher/description.md) and [simple-cipher](https://github.com/exercism/problem-specifications/blob/master/exercises/simple-cipher/description.md) seem similar enough that it's probably not worth having both.. Any thoughts on that, @kytrinyx, @petertseng?

petertseng:
https://github.com/exercism/problem-specifications/issues/605

dkinzer:
Ah, the post in that spec issue hits the nail on the head.

There are other tests that need to be included in order to really test this cipher; including the decryption which we can get by passing the negative key used to encrypt (i.e. if I encrypt by rotating all letter x steps forwards, then rotating all letters x steps backwards undoes the encryption).

Also one could add a rotating backwards beyond the first key, and probably also rotating x steps backwards greater than 26.

dkinzer:
@yurrriq hmmm clojure doen't have that simple cipher exercise yet, but it looks to me like they are very different.  This test case for the simple cipher reads:

```c#
Assert.Equal("zabcdefghi", cipher.Encode("zzzzzzzzzz"));
```

But the solution for the ROT cipher would be a single letter string.

```clojure
(assert (= (rotate "zzzzzzzzzz" 13) "mmmmmmmmmm"))
```

yurrriq:
Ah, ok. This LGTM. Shall I merge then?

dkinzer:
I think I've covered all the edge cases for backwards rotation, so this is good to merge imho.

yurrriq:
Cool. Can you add it to `config.json`, please?

yurrriq:
Thanks!

dkinzer:
Thanks!

PR #181, opened 2017-08-22
Update tests to reflect master problem spec

tstirrat15:
## Motivation
The [problem spec](https://github.com/exercism/problem-specifications/blob/master/exercises/pangram/canonical-data.json) for Pangram specifically indicates that the characters will be all ASCII. Additionally, there was one test from the spec missing.

## Changes
* Removed non-ascii test
* Added test for upper and lowercase characters not being counted separately


yurrriq:
With digging too deep this LGTM.

Issue #182, opened 2017-08-24
Clock tests weren't cleaned up?


tstirrat15:
[See here](https://github.com/exercism/clojure/blob/master/exercises/clock/clock.mustache).

I can see that that template was used to generate tests, but it wasn't cleaned up afterwards, which is potentially confusing to users.

Is there some cleanup process that didn't run?

yurrriq:
I don't follow. Could you please describe what, specifically, needs to be cleaned up?

tstirrat15:
The .mustache file. Isn't that a vestige of the build process that shouldn't be around once the tests are built?

yurrriq:
Ah, see. :+1: to excluding it from the user-facing side.

tstirrat15:
Should the build process automatically delete these? I couldn't grok how the build process works and wanted that to be fixed if it needed to be fixed. Otherwise I'm happy to submit a PR that just deletes all of the `.mustache`s from this repo.

yurrriq:
Deleting the mustache files is definitely incorrect. We just need to move them or otherwise prevent users from seeing them with they fetch from the Clojure track.

yurrriq:
@kytrinyx, any advice here? I know we've used `_src/` et al. before.

haus:
Yea, we're already using `_src`, so the templates could just move there and the generator be updated to account for the change. (unless there's an easy way to exclude the mustache files from being fetched).

kytrinyx:
We have a couple of ways of making sure things don't get fetched.

The first (and probably cleanest) is to stick everything you don't want fetched into a `.meta/` directory within the exercise directory. So `$ROOT/exercises/clock/.meta/$FILE`.

The other is to add a key `ignore_pattern` to the `config.json` (see https://github.com/exercism/docs/blob/master/language-tracks/configuration/README.md) which matches anything that should not be served to the user.

The `.meta/` directory would be a great place to put any build artifacts as well as the example solution.

haus:
@tstirrat15 i think this should be resolved with the linked PR, so i'm going to close it for now. if it is still an issue, please reopen it.

tstirrat15:
Looks good to me. Thank you!

PR #183, opened 2017-09-12
Issue 182: Update test generator to use .meta directory

haus:
Previously mustache files were in the main directory for each exercise,
which would cause them to be fetched for users, which can be confusing.
This commit updates the generator to use the .meta directory in each
exercise instead, which will not be fetched.


haus:
I think this should take care of #182

yurrriq:
LGTM

PR #184, opened 2017-09-19
adding collatz-conjecture

Average-user:
It was on the Unimplemented exercises of [exercism.io](http://exercism.io) so i decided to implement it.

### DIFF
--- a/exercises/collatz-conjecture/project.clj
+++ b/exercises/collatz-conjecture/project.clj
@@ -1,5 +1,4 @@
 (defproject collatz-conjecture "0.1.0-SNAPSHOT"
   :description "collatz-conjecture exercise"
-  ;;:url "https://github.com/exercism/xclojure/tree/master/exercises/collatz-conjecture"
-  :dependencies [[org.clojure/clojure "1.8.0"]
-                 [proto-repl "0.3.1"]])
+  :url "https://github.com/exercism/xclojure/tree/master/exercises/collatz-conjecture"
+  :dependencies [[org.clojure/clojure "1.8.0"]])
--- a/exercises/collatz-conjecture/project.clj
+++ b/exercises/collatz-conjecture/project.clj
@@ -1,0 +1,5 @@
+(defproject collatz-conjecture "0.1.0-SNAPSHOT"
+  :description "collatz-conjecture exercise"
+  ;;:url "https://github.com/exercism/xclojure/tree/master/exercises/collatz-conjecture"
+  :dependencies [[org.clojure/clojure "1.8.0"]
+                 [proto-repl "0.3.1"]])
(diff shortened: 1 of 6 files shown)


exercises/collatz-conjecture/project.clj
```diff
@@ -0,0 +1,5 @@
+(defproject collatz-conjecture "0.1.0-SNAPSHOT"
+  :description "collatz-conjecture exercise"
+  ;;:url "https://github.com/exercism/xclojure/tree/master/exercises/collatz-conjecture"
+  :dependencies [[org.clojure/clojure "1.8.0"]
+                 [proto-repl "0.3.1"]])
```
haus: The `proto-repl` dependency should get pulled out I think.

haus:
@Average-user thanks for much for the contribution. This looks like a great start.

exercises/collatz-conjecture/src/example.clj
```diff
@@ -0,0 +1,8 @@
+(ns collatz-conjecture)
+
+(defn collatz [n]
+  (when (> n 0)
+    (loop [t 0 n n]
```
haus: Style nitpick: I would generally put each pair of bindings on their own line to disambiguate them.

Average-user: i would if there were more than 2, but if you insist i will.

Average-user:
I did like this for  the Haskell repository and there were no problems, not yet at least.

But again ... If you insist , i can change it.

Average-user:
Do you know any other exercise that handle with errors ? i'm not completely sure of how would i do it, if i have to.

haus:
It looks like https://github.com/exercism/clojure/blob/4562cf14857a2b886e27344c2e5e56e20e7f99bc/exercises/wordy/src/example.clj#L16-L23 throws an exception, and the tests validate that an exception was thrown (https://github.com/exercism/clojure/blob/4562cf14857a2b886e27344c2e5e56e20e7f99bc/exercises/wordy/test/wordy_test.clj#L47-L55).

Average-user:
ok thanks.

exercises/collatz-conjecture/src/example.clj
haus: Looking at the other exercises, loop bindings are either on their own lines per pair, or comma separated if there are few. So if you wanted to add a comma between the two pairs, that would also be fine.

haus:
Thanks again @Average-user !

Average-user:
Should i do more?

Average-user:
I think that the source link of the README.md isn't working on the merge

haus:
If you find some other unimplemented exercises that you want to do, we'd love to have them.

I'll try to see what's up with the README.

Average-user:
I've send another pull request with another exercise. But i have no rush.

PR #185, opened 2017-09-20
Fixup README link in collatz-conjecture

haus:


haus:
@Average-user I think this fixes the README issue

Average-user:
Yeah i think so

PR #187, opened 2017-09-23
Implementing spiral-matrix

Average-user:
Well, the only concern about this that i have, is that the algorithm that i implemented is a translation of the one in here [http://rosettacode.org/wiki/Spiral_matrix#Common_Lisp](http://rosettacode.org/wiki/Spiral_matrix#Common_Lisp) that it is in Common Lisp. And i couldn't find anything about if its allowed or not , but on the other hand , is not even the same code is just a translation.


haus:
Sorry it's taken me a bit to get to this. I plan on looking it over this weekend.

Average-user:
@haus Is better like this, pls don't merge yet, I'm trying some stuff

haus:
@Average-user cool, let me know when you're ready and i'll look it over

Average-user:
@haus  But. Do you think that what I wrote about rosettacode might be a problem?

haus:
The code there is licensed under the [GNU Free Documentation License](https://www.gnu.org/licenses/fdl-1.2.html). I'll have to read that to see what it means for derivative works.

haus:
In general the GPL family of licenses are incompatible with the MIT license, which this repo is licensed as.

Average-user:
@haus Im not changing anything so, merge if you want

haus:
@yurrriq do you have any experience with algorithms derived from other works in pull requests?

Average-user:
@haus [http://www.gnu.org/licenses/fdl-1.2.html](http://www.gnu.org/licenses/fdl-1.2.html) Here it says this: 

*This License is a kind of "copyleft", which means that derivative works of the document must themselves be free in the same sense. It complements the GNU General Public License, which is a copyleft license designed for free software.*

And I think this project satisfy that, I think ..

yurrriq:
@haus I generally avoid such issues or try to contact the author..

yurrriq:
Perhaps we could make this more “Clojurian” (e.g. using transducers) and thereby feel it’s “different enough” (on top of being a different language to begin with). Something to consider is that there are only so many ways to implement such a problem, and solutions in Lisp-like languages are bound to look somewhat similar. This feels like a solidly grey area to me, but I think we can “lighten” it a bit, as suggested above.

Average-user:
That could work

haus:
thanks @Average-user !

PR #194, opened 2017-10-26
added run length encoding exercise

themistoklik:


exercises/run-length-encoding/project.clj
```diff
@@ -0,0 +1,5 @@
+
```
haus: I think this leading newline can be removed.

exercises/run-length-encoding/src/example.clj
```diff
@@ -0,0 +1,22 @@
+(ns run-length-encoding)
+
+(defn run-length-encode
+  ;encodes a string with run-length-encoding
```
haus: These comments can be literal strings. In this function it can be:

```clojure
(defn run-length-encode
  "encodes a string with run-length-encoding"
  [s]
```

haus: Same is true for the other functions in the namespace.

exercises/run-length-encoding/src/run_length_encoding.clj
```diff
@@ -0,0 +1,9 @@
+(ns run-length-encoding)
+
+(defn run-length-encode
+  ;encodes a string with run-length-encoding
```
haus: same comment about docstring applies here: they can be strings and not comments.

also, adding a parameter list to the stub functions would be good.

exercises/run-length-encoding/src/example.clj
```diff
@@ -0,0 +1,22 @@
...
+  [s]
+  (apply str
+         (for
+           [ x (partition-by identity s)]
+           (str
+             (if
```
haus: It looks like this `if` can be updated to use the `when` form instead as the if returns nil in once case.

`(if (= 1 (count x)) nil (count x))` can be rewritten as `(when-not (= 1 (count x)) (count x))`.

exercises/run-length-encoding/src/example.clj
```diff
@@ -0,0 +1,22 @@
...
+               (= 1 (count x)) nil (count x)) (first x)))))
+
+(defn run-length-decode
+  ;decodes a run-length-encoded string
+  [s]
+  (->> s
+       (re-seq #"[0-9]+.|[a-zA-Z\s]")
+       vec
```
haus: This `vec` call shouldn't be necessary.

exercises/run-length-encoding/src/example.clj
```diff
@@ -0,0 +1,22 @@
...
+       (mapv #(if (= (count %) 1) (str "1" %) %))
```
haus: These `mapv` calls can be `map` calls instead.

exercises/run-length-encoding/src/example.clj
```diff
@@ -0,0 +1,22 @@
...
+       (mapv #(vector (Integer. (reduce str "" (butlast %))) (str (last %))))
```
haus: This `vector` call can be replaced by a call to `list`.

haus:
@themistoklik docstrings come before the param list for functions. other than that it is looking great.

themistoklik:
Hey @haus , thank you for the comments.

Checked the travis.yml in the travis linter, seems valid. Anything else to check on my end?

config.json
```diff
@@ -28,6 +28,14 @@
     {
       "core": false,
       "difficulty": 1,
+      "slug": "run-length-encoding",
+      "topics": null,
+      "unlocked_by": null,
+      "uuid": "1efc777a-055f-0480-2251-39384d1c3460cd267e4"
```
haus: This UUID looks a little long (they are 32 hex digits + 4 hypens long), yours is 43 total in length. I'm going to replace it with a different UUID and merge this in.

haus:
@themistoklik looks like travis just had a hiccup earlier. rerunning took care of it. I pushed out an update to the UUID used, and I'm all good here.

haus:
Thanks again @themistoklik !

Issue #199, opened 2017-12-17
src folder missing with beer-song exercise


heyarne:
Hey there,
somehow `exercism fetch` fetched the beer-song exercise without a `src` folder. I head to create `src` and `beer_song.clj` myself because emacs didn't find them. Can I somehow help fix this?

haus:
@heyarne we have an open issue about adding stubs to all exercises: https://github.com/exercism/clojure/issues/144

If you wanted to open a PR adding beer_song.clj with an empty function in it, that would help the next person.

haus:
I tried this out again and the merged PR has addressed this, so I'm going to close it out. Thanks again @heyarne

PR #200, opened 2017-12-18
Add beer_song.clj stub

heyarne:
As discussed in #199.

The stub is a bit more... fleshed out than the other ones, but I guess that's okay given that it's basically the minimal amount of code to run the test suite without exceptions.


haus:
Nice. Thanks @heyarne ! This looks good. One thought I had was maybe moving the comment into docstrings for the functions? Something like this...

```clojure
(defn verse
"Returns the nth verse of the song"
[n])
```

heyarne:
Good idea, just did that.

haus:
@heyarne awesome. looks great!

PR #202, opened 2017-12-19
reverse-string: add to track

sjwarner-bp:
Add reverse-string to track as per [todo](http://exercism.io/languages/clojure/todo)


exercises/reverse-string/src/example.clj
```diff
@@ -0,0 +1,6 @@
+(ns reverse-string)
+
+(require '[clojure.string :as s])
+
+(defn reverse-string 
+  ([word] (str (s/reverse word))))
```
haus: I don't think you need to call `str` on the reversed string.

```
user=> (clojure.string/reverse "abc")
"cba"
```

exercises/reverse-string/src/example.clj
```diff
@@ -0,0 +1,6 @@
+(ns reverse-string)
+
+(require '[clojure.string :as s])
```
haus: Outside of a REPL, requires are generally included as part of the ns call. There is an example of how to add a require to a ns call here: https://clojuredocs.org/clojure.core/ns#example-542692cac026201cdc326b83

haus:
Looks great @sjwarner-bp ! I left a couple of small comments and I'm happy to merge this in after those are addressed.

haus:
In general, I think exercism exercises strive to avoid the built-in functions for problems such as these. I wonder what this would look at if `clojure.string/reverse` weren't used?

exercises/reverse-string/src/example.clj
sjwarner-bp: Thanks for letting me know that - I didn't realise this at first! :slightly_smiling_face:

sjwarner-bp:
That's an interesting point - in the Java track the `reverse()` method is used (which is why I used a built-in function here), but maybe it should be broken down into smaller character manipulation.

I think there are two ways to view it - if this is an early exercise, maybe newbies haven't used the `clojure.string` library yet, in which case I think that it is a good learning experience for them. Equally, if this is a little later in the track then it might make sense to force users to implement their own version of the function.

Let me know what you think - if you decide on the second option I am more than happy to change it up :+1:

haus:
@sjwarner-bp that makes sense to me. I'm in.

kytrinyx:
@sjwarner-bp I did some digging. This is the PR that removed two-fer. Was it unintentional?

sjwarner-bp:
Oops, that's my mistake - I didn't realise this would delete `two-fer`! Completely unintentional. How would you suggest is best for me to fix this? 🙂

haus:
oh weird. i saw the commit but it didn't show up in the diff.

@sjwarner-bp i think you should be able to revert your revert commit and PR that?

sjwarner-bp:
Yeah, my thoughts too - not sure why it's caused this when it wasn't in the diff! Maybe because it wasn't squashed?

I'll get on a PR.

haus:
cool. thanks @sjwarner-bp

kytrinyx:
not sure why it's caused this when it wasn't in the diff!

Merging a PR applies all the commits in the branch. The commit that added two-fer was already applied, so didn't get applied again, but the second commit, the one that removed it, did get applied.

Usually in this sort of situation you'd want to either wait for the first PR (two-fer) to get merged onto master, and then rebase onto the recent master, or merge master into the branch. That will remove the two-fer from the diff without actually applying the patch. Or you could rebase your branch and delete the original `two-fer` commit to remove it from the diff entirely.

sjwarner-bp:
Thanks for letting me know that @kytrinyx - I'll keep that in mind for the future 🙂

kytrinyx:
@sjwarner as you can imagine, I learned this the hard way 😂

PR #203, opened 2017-12-20
say: add to track

sjwarner-bp:
Add `say` exercise to clojure track

### DIFF
--- a/exercises/say/src/example.clj
+++ b/exercises/say/src/example.clj
@@ -1,10 +1,11 @@
 (ns say
-    (:require [clojure.pprint :as pp]))
+    (:require [clojure.pprint :as pp])
+    (:require [clojure.string :as s]))
 
 (defn number [input]
   (if
     (or (< input 0) (> input 999999999999))
-    (throw (Exception. "Out of range"))
-    (clojure.string/replace (clojure.pprint/cl-format nil "~R" input) #"," "")
+    (throw (IllegalArgumentException. "Out of range"))
+    (s/replace (pp/cl-format nil "~R" input) #"," "")
   )
 )
--- a/exercises/say/src/example.clj
+++ b/exercises/say/src/example.clj
@@ -1,0 +1,10 @@
+(ns say
+    (:require [clojure.pprint :as pp]))
+
+(defn number [input]
+  (if
+    (or (< input 0) (> input 999999999999))
+    (throw (Exception. "Out of range"))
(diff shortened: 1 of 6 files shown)


exercises/say/src/example.clj
```diff
@@ -0,0 +1,10 @@
+(ns say
+    (:require [clojure.pprint :as pp]))
+
+(defn number [input]
+  (if
+    (or (< input 0) (> input 999999999999))
+    (throw (Exception. "Out of range"))
```
haus: For input validation, I think you probably want to throw `IllegalArgumentException` or use a `:pre` assertion on the `number` function (for more details about `:pre`, here's a link: https://clojure.org/reference/special_forms#__code_fn_code_name_em_param_em_em_condition_map_em_em_expr_em)

haus:
Thanks for the submission @sjwarner-bp !

exercises/say/src/example.clj
sjwarner-bp: Thanks for the review @haus! I think that an IllegalArgumentException makes the most sense here, so that's what I've implemented. Let me know what you think :slightly_smiling_face:

exercises/say/src/example.clj
```diff
@@ -0,0 +1,11 @@
+(ns say
+    (:require [clojure.pprint :as pp])
+    (:require [clojure.string :as s]))
```
haus: Generally multiple requires are passed to the same require call, so this would be (https://clojuredocs.org/clojure.core/ns#example-542692cac026201cdc326b84):

```clojure
(:require [clojure.pprint :as pp]
          [clojure.string :as s])
```

exercises/say/src/example.clj
```diff
@@ -0,0 +1,11 @@
...
+
+(defn number [input]
+  (if
+    (or (< input 0) (> input 999999999999))
+    (throw (IllegalArgumentException. "Out of range"))
+    (s/replace (pp/cl-format nil "~R" input) #"," "")
```
haus: Style nitpick: usually closing parens are left on the last line of code, so this would be `"")))`

haus:
I left two style comments. If you feel like you'd like to tweak them that's awesome, otherwise I'm happy to merge this in.

exercises/say/src/example.clj
sjwarner-bp: Cool! I didn't know this - I'll fix it up :+1:

Issue #144, opened 2016-10-16
Add/require stubs for all exercises


canweriotnow:
Re conversation on #121 - make stubs a standard feature of clojure track projects... this is actually bringing clojure track closer to typical for the other Lisp family tracks, which IIRC provide basic structure templates for solutions.

TODO:
- [x] Stub all existing exercises
- [ ] Add stub instructions to CONTRIBUTING guide
- [ ] Add test-for-presence-of-stub to test suite.
- [ ] intelligent stub generator script for new exercises in `bin/` (maybe?)

haus:
I broke out the first checklist item into #205

PR #204, opened 2017-12-25
Add stub for hamming distance

heyarne:


heyarne:
Hm. So the further I progress, the less stubs are there. I guess the clean solution would really be to implement a stub generator, as discussed in #144?

haus:
@heyarne either a stub generator or a linter for new exercises that would fail if stubs don't exist.

Issue #206, opened 2017-12-27
Update to clojure 1.9.0


haus:
Now that clojure 1.9.0 has been released we should probably update all exercises to use it.

haus:
done.

PR #207, opened 2017-12-27
Update to clojure 1.9.0

haus:
This is a fairly painless update to 1.9.0. The only required change for
some exercises was moving from the string require to the keyword
:require to allow the `ns` calls to conform to the newly added spec.


haus:
Addresses #206

Issue #205, opened 2017-12-27
Stub all existing exercises


haus:
Here is a list of all of the exercises that currently exist:

- [x] accumulate
- [x] acronym
- [x] all-your-base
- [x] allergies
- [x] anagram
- [x] atbash-cipher
- [x] bank-account
- [x] beer-song
- [x] binary-search-tree
- [x] binary-search
- [x] binary
- [x] bob
- [x] bracket-push
- [x] change
- [x] clock
- [x] collatz-conjecture
- [x] complex-numbers
- [x] crypto-square
- [x] diamond
- [x] difference-of-squares
- [x] etl
- [x] flatten-array
- [x] gigasecond
- [x] grade-school
- [x] grains
- [x] hamming
- [x] hello-world
- [x] hexadecimal
- [x] isbn-verifier
- [x] isogram
- [x] kindergarten-garden
- [x] largest-series-product
- [x] leap
- [x] luhn
- [x] meetup
- [x] minesweeper
- [x] nth-prime
- [x] nucleotide-count
- [x] octal
- [x] pangram
- [x] pascals-triangle
- [x] perfect-numbers
- [x] phone-number
- [x] pig-latin
- [x] pov
- [x] prime-factors
- [x] protein-translation
- [x] proverb
- [x] queen-attack
- [x] raindrops
- [x] reverse-string
- [x] rna-transcription
- [x] robot-name
- [x] robot-simulator
- [x] roman-numerals
- [x] rotational-cipher
- [x] run-length-encoding
- [x] say
- [x] scrabble-score
- [x] secret-handshake
- [x] series
- [x] sieve
- [x] space-age
- [x] spiral-matrix
- [x] strain
- [x] sublist
- [x] sum-of-multiples
- [x] triangle
- [x] trinary
- [x] word-count
- [x] wordy

CrowsVeldt:
I'd be happy contribute. Is there more to it than adding an {exercise-name}.clj file with the appropriate namespace and functions declerations?

CrowsVeldt:
I believe all exercises have stub files with PR #233, although I think there's still room for improvement in terms of consistency and clarity.

CrowsVeldt:
I've done another pass and caught to exercises I missed previously, as well as editing the stub files for greater consistency.

haus:
Nice. This looks all done thanks @CrowsVeldt !

PR #208, opened 2017-12-28
Add stubs for exercises

haus:


haus:
Currently includes exercises starting with a through e. Will eventually resolve #205

haus:
With these stubs, tests will run but fail with arity errors, which should hopefully be enough to go on.

Issue #215, opened 2018-02-18
Hello, World! exercise has an old ReadMe


christianpoveda:
The readme just says you have to implement a function that returns the string "Hello, World!". However, some test ask for a function with parity 1 returning "Hello, <1st argument>"!

christianpoveda:
I could update it if you want and do a PR

haus:
@christianpoveda that would be awesome

PR #216, opened 2018-02-18
Fix hello world

christianpoveda:
This fixes https://github.com/exercism/clojure/issues/215


haus:
Looks great! Thanks so much @christianpoveda

PR #217, opened 2018-02-24
Fix 3-digit armstrong test description

imtayadeway:
The description contradicts what is being tested and needs to be in
the negative.

:heart:


haus:
Good catch. Thanks!

PR #218, opened 2018-03-18
add poker

Average-user:


haus:
Thanks for the PR @Average-user ! I'm going to start looking this over this weekend.

exercises/poker/src/example.clj
```diff
@@ -0,0 +1,42 @@
...
+  (:require [clojure.string :as cs]))
+
+(defn times [x xs]
+  (count (filter #{x} xs)))
+
+(defn rank-hand [hand]
+  (let [h   (cs/split (cs/replace hand #"10" "T") #" ")
+        r   (map #(.indexOf (seq "..23456789TJQKA") (first %)) h)
+        nr  (distinct r)
+        gs  (reverse (sort-by vec (map vector (map #(times % r) nr) nr)))
```
haus: I think if you used `frequencies` (https://clojuredocs.org/clojure.core/frequencies) here you wouldn't need the `times` function or the `nr` var.

`(reverse (sort-by vec (frequencies nr)))` returns a very similar structure to the current `gs`, except that the count is returned in the second position instead of the first.

exercises/poker/src/example.clj
```diff
@@ -0,0 +1,42 @@
...
+        cs  (mapv first gs)
```
haus: Maybe `rank-counts` instead of `cs`?

exercises/poker/src/example.clj
```diff
@@ -0,0 +1,42 @@
...
+        rs  (if (= (mapv second gs) [14,5,4,3,2]) [5,4,3,2,1] (mapv second gs))
```
haus: Maybe `normalized-ranks` instead of `rs`?

exercises/poker/src/example.clj
```diff
@@ -0,0 +1,42 @@
...
+          :else            [0 rs])))
+
+(defn greater-than [xs ys]
+  (or (> (first xs) (first ys))
+      (->> (map compare (second xs) (second ys))
+           (take-while (partial not= -1))
+           (some (partial = 1)))))
+
+(defn best-hands [hands]
+  (loop [m [0 []], r [], hs hands]
```
haus: I think `hands-remaining` might be a better var name than `hs`; `r` could be `results`; `m` could be `max-rank`.

exercises/poker/src/example.clj
```diff
@@ -0,0 +1,42 @@
...
+    (if (empty? hs)
+      r
+      (let [[x xs] ((juxt first rest) hs)
```
haus: I think you can get rid of the `(juxt first rest)` call here by rewriting this as `(let [[x & xs] hs]` to use some destructuring.

PR #219, opened 2018-04-17
add dominoes

Average-user:


haus:
Thanks for the PR @Average-user ! I'll try to dig in this week and the coming weekend.

exercises/dominoes/test/dominoes_test.clj
```diff
@@ -0,0 +1,39 @@
...
+  (is ((comp not can-chain?) [[1 2]])))
+
+(deftest three-elements
+  (is (can-chain?  [[1 2] [3 1] [2 3]])))
+
+(deftest can-reverse-dominoes
+  (is (can-chain? [[1 2] [1 3] [2 3]])))
+
+(deftest cant-be-chained
+  (is ((comp not can-chain?) [[1 2] [4 1] [2 3]])))
```
haus: The `(comp not can-chain?)` makes sense if lifted out into it's own `def`, but I think otherwise it is better to just write `(not (can-chain? ...))`

exercises/dominoes/src/example.clj
```diff
@@ -0,0 +1,22 @@
...
+          (= x (first xs)) (reduce conj ac (rest xs))
+          :else            (recur (rest xs) (conj ac (first xs))))))
+
+(defn connects [[a b] [c d]]
+  (cond (= b c) [c d]
+        (= b d) [d c]
+        :else   nil))
+
+(defn backtrack [rem chain]
+  (or (and (empty? rem) (= (ffirst chain) (second (peek chain))))
```
haus: I think using `last` here instead of `peek` will be more readable.

exercises/dominoes/src/example.clj
```diff
@@ -0,0 +1,22 @@
...
+      (and (not (empty? rem))
+           (some #(let [c (connects (peek chain) %)]
```
haus: I think this `(some...)` might do well to be lifted into its own function, and I think the same comment about `last` instead of `peek` applies.

exercises/dominoes/src/example.clj
```diff
@@ -0,0 +1,22 @@
+(ns dominoes)
+
+(defn rm [xs x]
```
haus: I think this function can be replaced with `(vec (remove #(= x) xs))`.

Average-user: The idea is to just remove one occurrence

exercises/dominoes/src/example.clj
Average-user: I was trying to avoid the use of  ``` declare ```.

exercises/dominoes/src/example.clj
haus: Ah. okay, that makes sense. I think a solution that avoided looping might be better and make it more obvious what is going on.

As it is currently written it took me a while to figure out what was going on and even then I missed a key part of what it was trying to do.

haus: As this function removes the first occurrence, a solution that found the first index of an element (maybe `.indexOf`) and then removed that element, would probably be good.

exercises/dominoes/src/example.clj
yurrriq: From the `peek` docstring:
> ... for a vector, same as, but much more efficient than, `last`.

Average-user: @yurrriq  That was my reason originally, but I'm not sure if is really worth it

exercises/dominoes/src/example.clj
Average-user:
```Clojure
(defn rm [xs x]
  (let [i (.indexOf xs x)]
    (if (= -1 i)
      xs
      (reduce conj (subvec xs 0 i) (subvec xs (inc i) (count xs))))))
```
Something like this?

exercises/dominoes/src/example.clj
haus: Yea, my comment was mainly just that without reading the docs I wouldn't expect `peek` to have that behavior, so `last` is more readable (and matches better with the `ffirst` call). I don't feel super strongly.

exercises/dominoes/src/example.clj
haus: yea, that seems better to me.

haus:
Looking good. I had one overall question, which is whether the dominoes solution and test should be generating a valid solution, or just saying that it is solvable. It seems like both of these already exist in exercism (the java dominoes solution does what this does, which is to decide if the chain is solvable, while the ruby solution actually generates a valid solution).

Average-user:
I could do that too, I just saw that the tests were just to ccheck if it is possible, maybe they changed meanwhile

haus:
I'm in. thanks for the PR @Average-user !

PR #220, opened 2018-04-28
Update minesweeper tests for windows

haus:
The solution for minesweeper uses cl-format, which will emit the correct
line separator for a given platform. The tests used \newline, which is
always \n, and not \r\n. This causes tests to fail on windows, so this
commit updates the tests to use the line separator for the platform
instead of just \n.


yurrriq:
LGTM! Should we set up [AppVeyor](https://www.appveyor.com/)?

yurrriq:
Here's the (likely bit-rotten) [LFE config](https://github.com/exercism/lfe/blob/master/appveyor.yml), incase that's useful.

haus:
Appveyor sounds good to me. I'll see what i can put together.

haus:
Appveyor setup in https://github.com/exercism/clojure/pull/222

PR #221, opened 2018-04-28
Add stubs for more exercises

haus:
This PR adds stubs for more of the existing exercises to help folks get up and running.


exercises/gigasecond/src/gigasecond.clj
```diff
@@ -0,0 +1,5 @@
+(ns gigasecond)
+
+(defn from []  ;; <- arglist goes here
+    ;; your code goes here
+)
```
yurrriq: I like a newline at the end of a file (applies to all but minesweeper.clj in this PR), but this looks good to me otherwise.

haus: cool, adding newlines is easy enough. i'll push that out.

PR #224, opened 2018-05-05
Add appveyor badge to README.org

haus:


haus:
Not sure if it is worth pinning the badge to the master branch status or the general status. Travis seems pinned to general status.

yurrriq:
Nice! I have a _mild_ preference for master, since I believe that's what ships.

PR #226, opened 2018-05-22
Say - Add referance to espeak

CrowsVeldt:
I really liked the challenge to use text-to-speech software to say the numbers out loud, but since the
README says "for Mac Users" I skimmed past it at first without paying attention. I'm glad I took
the time to find an alternative, and I wanted to make sure others will be made aware of it too, so I added a reference to espeak for Windows and Linux users.

### DIFF
--- a/exercises/say/README.md
+++ b/exercises/say/README.md
@@ -24,7 +24,8 @@ Some good test cases for this program are:
 ### Extension
 
 If you're on a Mac, shell out to Mac OS X's `say` program to talk out
-loud.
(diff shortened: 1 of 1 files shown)


haus:
@CrowsVeldt thanks for the PR! I wonder if you'd also be willing to make a PR to https://github.com/exercism/problem-specifications/blob/master/exercises/say/description.md which is the original source for this README?

CrowsVeldt:
I'd be happy to  :)

PR #233, opened 2018-06-06
Issue #205: Add stub files

CrowsVeldt:
Added files for leap, luhn, meetup, nth-prime, nucleotide-count, octal, pangram, and pascals triangle.

I based the stub file for pascals triangle on the example file for that exercise, but I'm not sure it's clear enough for someone coming to it cold to understand. I'd be happy to take suggestions on how to improve it. (or any of them, to be clear)

### DIFF
--- a/exercises/accumulate/src/accumulate.clj
+++ b/exercises/accumulate/src/accumulate.clj
@@ -1,0 +1,5 @@
+(ns accumulate)
+
+(defn accumulate [] ;; <- Arglist goes here
+      ;; your code goes here
+      )
--- a/exercises/accumulate/src/accumulate.clj
+++ b/exercises/accumulate/src/accumulate.clj
@@ -1,5 +1,5 @@
 (ns accumulate)
 
-(defn accumulate [] ;; <- Arglist goes here
+(defn accumulate [] ;; <- arglist goes here
       ;; your code goes here
       )
--- a/exercises/armstrong-numbers/src/armstrong_numbers.clj
+++ b/exercises/armstrong-numbers/src/armstrong_numbers.clj
@@ -1,3 +1,5 @@
 (ns armstrong-numbers)
 
-(defn armstrong? [n])
+(defn armstrong? [n] ;; <- arglist goes here
+  ;; your code goes here
+  )
--- a/exercises/collatz-conjecture/src/collatz_conjecture.clj
+++ b/exercises/collatz-conjecture/src/collatz_conjecture.clj
@@ -1,3 +1,5 @@
 (ns collatz-conjecture)
 
-(defn collatz [n])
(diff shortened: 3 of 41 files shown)


CrowsVeldt:
Added stubs for perfect-numbers, phone-number, pig-latin, and pov.

Is the .gitignore file in exercises/pov still needed for something? It doesn't seem to be necessary as far as I can tell, but I'm not sure why it was added in the first place. I've removed it for now on the assumption that it's not needed. I'll revert that commit if I'm wrong about that.

CrowsVeldt:
Added stub files for prime-factors, protein-translation, queen-attack

CrowsVeldt:
I believe that's all the exercises stubbed. I plan on taking another pass to make sure everything is consistent.

CrowsVeldt:
Turns out I missed two exercises.

CrowsVeldt:
I edited all stubs to have the same format except for:

- `beer-song` and `run-length-encoding`, where I think argument names and doc strings do a better job of guiding the reader than `arglist goes here`, etc.

- `pascals-triangle` and `proverb`, where I'm not sure how to handle the non-function `def`s

CrowsVeldt:
Turns out I missed one. That'll teach me not to triple check.

On a more serious note: I think that the stub files for `beer-song` and `run-length-encoding` (where the functions have arguments and doc strings provided) are better than the others, at least for less experienced programmers. That way you get an overview of the exercise from the README, and the src file itself provides guidance as to expected input and output values.

I suppose there's an argument to be for the value of providing less direct guidance so that the user has to figure things out from the README and tests, but from what I've seen I think that overall Exercism leans towards being as welcoming as possible, even to people with no programming experience whatsoever. 

In any event. I'll double check the issues to see if a decision has already been reached, and if not I'll open an issue for it myself. I'll be happy to try to do more work on the stub files myself, but as I'm not terribly experienced it'll probably take some guidance and editing to get to something really good.

haus:
@CrowsVeldt Awesome work! I have no objection to adding args to some of the stubs (probably the earlier exercises in the sequence?), especially as they will allow tests to run and just fail instead of throwing exceptions about incorrect arity of functions. That could be a follow up PR, unless you wanted to keep working on this one.

CrowsVeldt:
Thanks! I'd be happy to do it either way, is there a standard procedure for when to start a new PR?
and what would you consider as an appropriate place to stop providing args?

haus:
Sorry for the delay in responding. I don't think there's a standard for when to start a new PR, and I don't know when is the right point to stop providing args, maybe a third or halfway through the set?

haus:
I'll merge this in for now

CrowsVeldt:
Cool! I guess I'll add args for the first third or so, and we'll see how it looks then

PR #239, opened 2018-06-16
Add an ISBN Verifier test case for short ISBN

jgomo3:
If your code verify the isbn length to be only *less than or equal to* 10, it will accept some ISBN with less than 10 digits which still satisfy other rules, being "00" a case which could be considered ISBN by some "almost correct" solutions missing this consideration.

I.e: http://exercism.io/submissions/f2fac1e79bf441e08afecd196ec0a442

### DIFF
--- a/exercises/isbn-verifier/test/isbn_verifier_test.clj
+++ b/exercises/isbn-verifier/test/isbn_verifier_test.clj
@@ -32,6 +32,9 @@
 (deftest too-long-isbn-and-no-dashes
   (is (= false (isbn? "3598215078X"))))
 
+(deftest too-short-isbn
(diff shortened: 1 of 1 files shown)


haus:
Thanks for the PR @jgomo3 ! I'm happy to add this as a test case here, I wonder if you'd be open to making a PR to the shared test cases for all exercism tracks here https://github.com/exercism/problem-specifications/blob/master/exercises/isbn-verifier/canonical-data.json as well?

jgomo3:
@haus Sure: https://github.com/exercism/problem-specifications/pull/1251

haus:
awesome thanks!

PR #240, opened 2018-06-18
Add default args to stub files

CrowsVeldt:
I added, or updated, default arguments in the first ~25 exercises' stub files. I tried to make the argument names input format and/or the scenario of the exercise, except in cases where it seemed completely redundant to do so (i.e. when a `num` really is just a `num`). Of course I'm open to feedback and will be happy to make changes as necessary.


haus:
Thanks for the PR @CrowsVeldt ! This looks great to me.

PR #244, opened 2018-07-20
collatz conjecture expect to use precondition instead of IllegalArgumentException

razum2um:
I think it's more clojurish


haus:
Thanks for the PR @razum2um ! I'm of mixed feelings when considering using a precondition instead of throwing. Mainly preconditions can't pass in a custom error message, and the error message that comes out isn't super great. For example:

Calling something like `(defn foo [n] {:pre [(pos? n)]} (println n))` with -1 results in 
`AssertionError Assert failed: (pos? n)  user$foo.invokeStatic (:1)` which isn't as clear as something like `foo expects n to be a positive integer`.

razum2um:
@haus got you 👍so updated spec to just allow both approaches

PR #249, opened 2018-08-08
Adding a test for the armstrong number 21897142587612075.

mawis:
I would like to propose this change from my experience in mentoring this exercise.

Many students seem to implement this exercise using floating point
arithmetics `(Math/pow)` comparing the result with the equal operator
afterwards.

I think we can agree, that comparing floating point numbers for strict
equality is not a good thing. But our tests had no case, where this
actually mattered.

I was searching for the lowest armstrong number, where this actually
matters and found 21897142587612075 to be a good candidate. The exercise
implemented in floating point arithmetics doesn't pass this added test,
while implementing it in integer arithmetics let all test pass.

### DIFF
--- a/exercises/armstrong-numbers/test/armstrong_numbers_test.clj
+++ b/exercises/armstrong-numbers/test/armstrong_numbers_test.clj
@@ -33,3 +33,7 @@
 (deftest not-armstrong-number-9926314
   (testing "Seven digit number that is not an Armstrong number"
     (is (not (armstrong? 9926314)))))
+
+(deftest armstrong-number-21897142587612075
+  (testing "Seventeen digit number that is an Armstrong number"
+    (is (armstrong? 21897142587612075))))


haus:
Thanks @mawis ! This seems like a reasonable test case to add. It might make sense to also open a PR against https://github.com/exercism/problem-specifications/blob/master/exercises/armstrong-numbers/canonical-data.json which is where shared test cases live (and some tracks generate the tests using this data).

haus:
Instead of `Math/pow` I assume the idea would be to use something like `(reduce * (repeat exp base))`?

haus:
Also it looks like the example solution here uses `Math/pow`, so maybe worth updating it as well (https://github.com/exercism/clojure/blob/42f9ab104d1e9291fe01265b379addeeee98c5cb/exercises/armstrong-numbers/src/example.clj).

mawis:
Good point, I actually didn't check the example solution.

I think `(reduce * (repeat exp base))` is a reasonable implementation. After asking the students not to use floating point arithmetics maybe half of them implement it like that (the other half uses clojure/math.numeric-tower).

haus:
I was surprised that the tests were still passing (as the example uses `Math/pow`), but I realized it was combined with `==`, so the tests still pass. So I don't think updating the example solution is strictly required.
So if you're satisfied with this as is, I'm happy to merge it in.

Issue #254, opened 2018-09-25
Proverb: README inconsistent with test


m1kal:
For `proverb` exercise the README states:

> Given a list of inputs, generate the relevant proverb. For example, given the list ["nail", "shoe", "horse", "rider", "message", "battle", "kingdom"], you will output the full text of this proverbial rhyme:
> 
> For want of a nail the shoe was lost.
> For want of a shoe the horse was lost.
> For want of a horse the rider was lost.
> For want of a rider the message was lost.
> For want of a message the battle was lost.
> For want of a battle the kingdom was lost.
> And all for the want of a nail.
> 

The test requires the last line to be "And all for the want of a horseshoe nail."

Aligning the test to exercise description would simplify the exercise and make the already submitted solutions seem too complex.

Maybe the README should be changed.

haus:
interesting. This matches the readme in the problem specifications repo, and the tests in that repo specify nail, not horseshoe nail. the wikipedia page shows both options (https://en.wikipedia.org/wiki/For_Want_of_a_Nail). updating the test here seems reasonable to me as the canonical source of truth references it without horseshoe (https://github.com/exercism/problem-specifications/blob/master/exercises/proverb/canonical-data.json).

PR #259, opened 2018-10-11
Proverb: align the test with README.

m1kal:
Solves #254.

The README shows last line as "And all for the want of a nail.".
The test requires the last line to be "And all for the want of a horseshoe nail.".


yurrriq:
LGTM, thanks!

PR #263, opened 2019-01-02
reformat code with cljfmt fix

AndreaCrotti:
I noticed a lot of code formatted very weirdly and I just ran a:
`for x in *; do cd $x && lein cljfmt fix && cd ..; done`

to fix it all in one go...
Ideally we could even enforce it the CI but maybe it's not even necessary if problems are not added very often.


haus:
Awesome thanks! I for one would not be opposed to a CI step to test for bad formatting, so if you'd like to set that up that would be rad.

PR #264, opened 2019-01-23
Redid Proverb Exercise

UnrelatedString:
Proverb's test suite currently tests a single imported variable `proverb` to see if it matches the full text of the proverb, and does nothing else. The template for writing a solution doesn't even suggest that you use a function of any sort, simply giving you `(def proverb )`. This is inconsistent with both the README and the canonical test data for the exercise. I have corrected this problem, and look forward to submitting my updated solution.

### DIFF
--- a/exercises/proverb/src/example.clj
+++ b/exercises/proverb/src/example.clj
@@ -1,15 +1,15 @@
 (ns proverb
   (:require [clojure.string :as str]))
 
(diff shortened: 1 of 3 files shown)


haus:
Thanks for the PR @UnrelatedString ! I should have a chance to look this over tonight and tomorrow.

Issue #253, opened 2018-09-20
clojure namespace as function name


pouya-abbassi:
In [nucleotide-count](https://github.com/exercism/clojure/tree/master/exercises/nucleotide-count) exercise, main function is called `count` which is a `clojure.core` function name. And you can't change it unless you change the tests.
As far as i know, you should avoid using the same names as `clojure.core` defines.

Is it a mistake or you want to teach how to use`:refer-clojure`?

yurrriq:
Nice catch! Seems like a good time to introduce `:refer-clojure` to me.

AndreaCrotti:
I would simply rename the function to something else imho

yurrriq:
Pull requests are most welcome!

haus:
Addressed with #273

Issue #228, opened 2018-05-31
Rename nucleotide-count/count to avoid shadowing clojure.core/count?


CrowsVeldt:
I don't imagine this is a major problem or anything, but maybe it'd avoid some inexperienced programmers (like myself) avoid confusion.

I'm not sure what a good alternative might be, maybe `amount`. Or maybe `count` could become `nucleotide-count` and the current `nucleotide-counts` could become `nucleotide-totals`.

CrowsVeldt:
[I forgot to add the url to the test file](https://github.com/exercism/clojure/blob/master/exercises/nucleotide-count/test/nucleotide_count_test.clj)

haus:
Addressed in #273

PR #273, opened 2019-05-21
fix function name collision exercism#253

pouya-abbassi:
Issue fix. #253

### DIFF
--- a/exercises/nucleotide-count/src/nucleotide_count.clj
+++ b/exercises/nucleotide-count/src/nucleotide_count.clj
@@ -1,8 +1,8 @@
 (ns nucleotide-count)
 
-(defn count [nucleotide strand] ;; <- Arglist goes here
+(defn counter [nucleotide strand]) ;; <- Arglist goes here
   ;; your code goes here
-)
+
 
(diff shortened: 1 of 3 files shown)


pouya-abbassi:
Also #228

haus:
Thanks for the PR @pouya-abbassi ! I'll take a look at this in the next couple of days.

exercises/nucleotide-count/src/nucleotide_count.clj
```diff
@@ -1,8 +1,8 @@
 (ns nucleotide-count)
 
-(defn count [nucleotide strand] ;; <- Arglist goes here
+(defn counter [nucleotide strand]) ;; <- Arglist goes here
   ;; your code goes here
-)
```
haus: would probably be good to leave this paren where it was. otherwise the comments about where to put code are a little misleading.

haus:
Looks good to me. If you have a chance to address the one comment I'm happy to merge it in.

exercises/nucleotide-count/src/nucleotide_count.clj
haus: i pushed out a commit to resolve this

PR #277, opened 2019-07-17
Add reverse-string tests

mtkp:
- single letter word test
- very long string test to encourage `reduce`, since a non-TCO recursive call will stack overflow.


haus:
These seem totally reasonable to add. It might be worth also adding them to https://github.com/exercism/problem-specifications/blob/master/exercises/reverse-string/canonical-data.json so that other tracks can also benefit from them (maybe not the second, as it requires function calls).

mtkp:
thanks @haus. i'll consider putting up a PR to the canonical test set.

PR #278, opened 2019-07-17
update armstrong reference impl

mtkp:
this brings the reference impl more inline with the mentor notes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,7 +2,7 @@
 language: clojure
 sudo: false
 lein: lein
-jdk: oraclejdk8
+jdk: openjdk8
 script:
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -2,7 +2,7 @@
 language: clojure
 sudo: false
 lein: lein
-jdk: oraclejdk8
+jdk: openjdk8
```
mtkp: switched to openjdk8 since this build now runs on xenial and there's an issue pulling oraclejdk8 from xenial. another option would be to configure the dist as trusty and keep oraclejdk8.

haus:
thanks for the PR @mtkp !

PR #281, opened 2019-11-20
Update tests for 'Leap' exercise

ovidiu141:
Added more tests for the 'Leap' exercise (based on the problem specifications)


haus:
awesome. thanks for the PR @ovidiu141 ! looks good from me at a first pass. i'll give it a deeper look this afternoon.

PR #284, opened 2019-11-25
Update tests for 'Triangle' exercise

ovidiu141:
Updated tests based on the problem specifications. Also updated the example.

### DIFF
--- a/exercises/triangle/src/example.clj
+++ b/exercises/triangle/src/example.clj
@@ -1,10 +1,17 @@
-(ns triangle
-  (:refer-clojure :exclude [type]))
+(ns triangle)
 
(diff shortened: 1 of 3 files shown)


haus:
@ovidiu141 Thanks for the PR! the new test cases look good.

i was wondering if you could elaborate a bit on the changes you made to the example and/or why you think the example is better now than it was before.

ovidiu141:
@haus The logic inside the test was good. I updated it in order to match the new structure of the exercise, so all new tests are passing for the example version.

porkostomus:
I see that the other language tracks have done likewise with this exercise, bringing it in sync with the new wording. This looks great, thank you!

PR #285, opened 2019-12-02
Updates armstrong-numbers to 1.1.0

amscotti:
Hi, this PR is an update to `armstrong-numbers` to update the unit test to match the [problem specifications](https://github.com/exercism/problem-specifications/blob/master/exercises/armstrong-numbers/canonical-data.json) version `1.1.0`

I'm still learning Clojure, so any suggestions would be helpful!

Thanks!


haus:
Looks great! thanks for the PR @amscotti

PR #286, opened 2019-12-02
Updates reverse-string to 1.2.0

amscotti:
Hi, this PR is an update to `reverse-string` to update the unit test to match the[ problem specifications](https://github.com/exercism/problem-specifications/blob/master/exercises/reverse-string/canonical-data.json) version 1.2.0

I'm still learning Clojure, so any suggestions would be welcomed!

Thanks!


haus:
Sorry for the delay in review. Looks great and thanks for the PR!

PR #292, opened 2020-02-11
Reorder "Armstrong numbers" to be later in the queue

JimLynchCodes:
Personally, I think that "armstrong numbers" is quite a difficult problem to be the very first one after "hello world"!

The first "real" problem in most other language tracks is "two-fer", and this seems about the right difficulty to me. 

I personally think the "reverse-string" and "bob" questions are a bit easier than "Armstrong numbers" as well which is why I have moved it down three spots below these three mentioned exercises. 

Go Clojureeeeeeeeeee! ❤️ 🔥 👍

### DIFF
--- a/config.json
+++ b/config.json
@@ -13,16 +13,6 @@
       "topics": null
     },
     {
-      "slug": "armstrong-numbers",
-      "uuid": "a4bdbdfd-1db1-425c-a243-dd032dc7b93a",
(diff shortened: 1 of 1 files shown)


JimLynchCodes:
bump :)

porkostomus:
Thanks a lot! The ordering is much better now.

PR #293, opened 2020-03-10
add docstrings for exercises A-E

porkostomus:
Bringing exercises up to date with our current working style-guide, which I've taken as far as providing docstrings for the function stubs.

It seems to me that they are very helpful and would expect students to appreciate them. Plus they are fun to write and I've been learning a lot by doing it. Which actually gives me some second thoughts -

Am I robbing the student of the experience of learning to write effective docstrings by doing this work for them? The main purpose was to encourage their use, but this could also be accomplished with a generic placeholder.

Feedback appreciated!

### DIFF
--- a/exercises/acronym/src/acronym.clj
+++ b/exercises/acronym/src/acronym.clj
@@ -1,5 +1,7 @@
 (ns acronym)
 
-(defn acronym [] ;; <- arglist goes here
-  ;; your code goes here
+(defn acronym
+    "Converts a phrase to its acronym."
+  [text]
+  ;; function body
 )
--- a/exercises/accumulate/src/accumulate.clj
+++ b/exercises/accumulate/src/accumulate.clj
@@ -1,5 +1,10 @@
 (ns accumulate)
 
-(defn accumulate [] ;; <- arglist goes here
-      ;; your code goes here
+(defn accumulate
+    "Given a coll and a function to perform
+   on each element of the coll, returns a new
+   coll containing the result of applying f
+   to each element of the input coll."
+  []
+      ;; function body
 )
--- a/exercises/accumulate/src/example.clj
+++ b/exercises/accumulate/src/example.clj
@@ -1,6 +1,11 @@
 (ns accumulate)
 
-(defn accumulate [f xs]
+(defn accumulate
+  "Given a coll and a function to perform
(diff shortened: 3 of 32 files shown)


exercises/acronym/src/acronym.clj
```diff
@@ -1,5 +1,7 @@
 (ns acronym)
 
-(defn acronym [] ;; <- arglist goes here
-  ;; your code goes here
+(defn acronym
+    "Converts a phrase to its acronym."
```
jr0cket: Would this be a more consistent doc string?

"Given a string containing a phrase,
convert the phrase to its acronym.
Return the acronym as a string."

porkostomus: Yes that's much better :)

It reads imperatively, however. I'm sure both ways are valid, but I thought it would be better to stick to the convention used in `clojure.core` docstrings which simply _describe_ what the function _does_.

jr0cket: Ah yes, looking at the style guide again reminded me why I do mine in a different order to the above.  So normally I would write

"Covert a phrase to it acronym

<optionally any additional explanation that  is useful, usually nothing for helper functions, but sometime some logic or business process that is useful to define>

Arguments: a string containing a phrase
Returns: acronym as a string"

`clojure.core` is not always a good guide to application level code.  That library was designed for very general functions and therefore has very little in the way of context.  Application code has context (except helper functions).  However, in small challenges this is less relevant.

As mentioned on Slack, I am still discovering what the main focus of exercism is, so am not suggesting a particular approach is correct, but think we should seek consistency as much as possible.

bemself:
> Am I robbing the student of the experience of learning to write effective docstrings by doing this work for them? The main purpose was to encourage their use, but this could also be accomplished with a generic placeholder.

@porkostomus I feel the same：-）

If I get it right, once Concept exercise is online, students will be asked to finish Concept before getting into this one. Since we have docStrings in each concept exercise, we can always give one sample of docstring and leave others to students to write their own. Then when they come to this practice exercise, they do themselves, that means we put a generic placeholder.

and before Concept exercise goes online, I presonally also would like to suggest a placeholder, but adding a sample effective docstring in the instruction README doc.

PR #301, opened 2020-11-23
Fixing phone-number test cases

Testare:
README for the phone-number exercise specifies certain phone number digits as being 2-9, but the tests allow and use 0-9 for these digits.

Perhaps validating these digits conform is out of scope for these tests, but the tests are also written so that people who want to add these validations actually fail the tests.

### DIFF
--- a/exercises/phone-number/src/example.clj
+++ b/exercises/phone-number/src/example.clj
@@ -7,7 +7,7 @@
 
 (defn- extract-parts
   [input]
-  (if-let [matches (re-find #"^1?(...)(...)(....)$" input)]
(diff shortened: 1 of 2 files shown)


haus:
@Testare looks like the example implementation might need updating to account for the test changes? i haven't looked too deeply into it yet

Testare:
Yes, sorry. This is my first time committing to an exercism repository, I didn't know about the example files.

I have updated the example file so that it passes the new tests.

haus (approved):
Looks great! Thanks for the PR @Testare

Issue #313, opened 2021-01-28
[v3] Implement new Concept Exercise: Character types


porkostomus:
This issue describes how to implement the `Character types` concept exercise for the Clojure track.

## Getting started

**Please please please read the docs before starting.** Posting PRs without reading these docs will be a lot more frustrating for you during the review cycle, and exhaust Exercism's maintainers' time. So, before diving into the implementation, please read up on the following documents:

- [The features of v3](https://github.com/exercism/v3/blob/master/docs/concept-exercises.md).
- [Rationale for v3](https://github.com/exercism/v3/blob/master/docs/rationale-for-v3.md).
- [What are concept exercise and how they are structured?](https://github.com/exercism/v3/blob/master/docs/features-of-v3.md)

## Goal

The goal of this exercise is to teach the student how character types are implemented in Clojure. According to the [official guide](https://clojure.org/guides/learn/syntax) this includes 3 main types: _strings_, _characters_ and _regular expressions_.

## Things to teach

This exercise aims to teach the student:

- Clojure strings are Java strings - eg. they are _immutable_
- Contained in double quotes
- Individual characters represented w/ leading backslash `\`
- Special characters (\newline \spec \tab, etc.)
- Unicode characters (\uNNNN or in octal with \oNNN)
- Literal regex syntax - `#"[0-9]+"`

## Things not to teach

Things outside the scope of this exercise are:

- Advanced regex usage

## Concepts

This exercise teaches the following concepts:

- `char`
- `string`
- `regular_expression`

## Prerequisites

This is a base-level concept and likewise only assumes general knowledge of English letters.

### Hints

- [Syntax guide](https://clojure.org/guides/learn/syntax) 

### After

- [java.util.regex.pattern](https://docs.oracle.com/javase/8/docs/api/java/util/regex/Pattern.html)
- [Clojure reference](https://clojure.org/reference/reader)
- [Java escape sequences](https://docs.oracle.com/javase/tutorial/java/data/characters.html)
## Representer

This exercise does not require any modifications to the track's representer.

## Analyzer

This exercise does not require any special analysis to be added for it.

## Implementing

To implement a concept exercise, the following files must be created:

```
languages
└── clojure
    └── exercises
        └── concept
            └── <SLUG>
                ├── .docs
                |   ├── instructions.md
                |   ├── introduction.md
                |   ├── hints.md
                |   └── after.md (optional)
                ├── .meta
                |   |── design.md
                |   └── Example.clj
[...]

PR #327, opened 2021-02-04
add squeaky-clean exercise

porkostomus:
Adds characters concept exercise, this PR replaces https://github.com/exercism/v3/pull/2830 and is connected to issue #313 .

Moved all practice exercises to practice folder.


exercises/concept/squeaky-clean/src/exercism/squeaky_clean.clj
```diff
@@ -0,0 +1,35 @@
+(ns exercism.squeaky-clean
```
ErikSchierboom, line 1: This looks like the exemplar file, in which case it should be located in the `.meta` directory.

ErikSchierboom (approved):
One minor comment. I would suggest merging this after https://github.com/exercism/clojure/pull/314 is merged today, otherwise there are a lot of unrelated changes.

PR #329, opened 2021-02-06
[v3] Add key features and tags

porkostomus:
Issues: #316 and #317.

### DIFF
--- a/config.json
+++ b/config.json
@@ -757,6 +757,37 @@
       "blurb": "TODO: add blurb for lists concept"
     }
   ],
-  "key_features": [],
+  "key_features": [
+    {
+      "title": "Lisp",
+      "content": "Clojure extends Lisp's code-as-data system beyond parenthesized lists (s-expressions) to vectors and maps, and has a powerful macro system.",
+      "icon": "features-lisp"
+    },
+    {
+      "title": "Dynamic",
+      "content": "Clojure's primary programming interface is the Read-Eval-Print-Loop (REPL). Code input into the REPL is compiled to JVM bytecode on the fly.",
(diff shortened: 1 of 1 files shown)


ErikSchierboom (approved):
As it looks like nobody else is reviewing, LGTM!

ErikSchierboom:
@porkostomus There are a couple of issues with the tags and key features according to configlet (must have been a very old version that checked the PR :)):

```
The value of `key_features.content` that starts with `Clojure extends Lisp's co`... is 139 characters, but it must not exceed 100 characters:
/home/erik/exercism/clojure/config.json

The value of `key_features.content` that starts with `Clojure's primary program`... is 140 characters, but it must not exceed 100 characters:
/home/erik/exercism/clojure/config.json

The value of `key_features.content` that starts with `Clojure provides the tool`... is 166 characters, but it must not exceed 100 characters:
/home/erik/exercism/clojure/config.json

The value of `key_features.content` that starts with `Clojure is impure, in tha`... is 133 characters, but it must not exceed 100 characters:
/home/erik/exercism/clojure/config.json

The value of `key_features.content` that starts with `Clojure, being a practica`... is 224 characters, but it must not exceed 100 characters:
/home/erik/exercism/clojure/config.json

The value of `key_features.content` that starts with `Clojure is designed to be`... is 234 characters, but it must not exceed 100 characters:
/home/erik/exercism/clojure/config.json

The `tags` array contains `used_for/used_for/frontends`, which is not one of the allowed values:
/home/erik/exercism/clojure/config.json
```

But those can be fixed in a follow-up PR :)

PR #369, opened 2021-06-13
Add two more diagonal test cases for queen attack

fernandrone:
The current diagonal tests take place in a situation where the white piece is "to the left and above" the black piece, which means that a technically invalid solution, which does not properly consider all diagonal attack vectors, may pass all test cases.

Given the current test cases:

```clojure
  (is (= true  (queen-attack/can-attack {:w [1 1] :b [6 6]})))
  (is (= true  (queen-attack/can-attack {:w [0 6] :b [1 7]})))
  (is (= true  (queen-attack/can-attack {:w [4 1] :b [6 3]})))
```

If we decompose `w` and `b` as `{[x1 y1] :w, [x2 y2] :b}`, then:

```
x1 <= x2
y1 <= y2
```

This commit adds two more valid diagonal test cases:

```clojure
  (is (= true  (queen-attack/can-attack {:w [2 3] :b [5 0]})))
  (is (= true  (queen-attack/can-attack {:w [2 3] :b [0 5]}))))
```

These will fail certain implementations which would pass the previous test suite, such as:

```clojure
(= (- y2 y1) (- x2 x1))
```

Or:

```clojure
(= (- x1 y1) (- x2 y2))
```

### DIFF
--- a/exercises/practice/queen-attack/test/queen_attack_test.clj
+++ b/exercises/practice/queen-attack/test/queen_attack_test.clj
@@ -33,4 +33,6 @@
   (is (= true  (queen-attack/can-attack {:w [5 4] :b [2 4]})))
   (is (= true  (queen-attack/can-attack {:w [1 1] :b [6 6]})))
   (is (= true  (queen-attack/can-attack {:w [0 6] :b [1 7]})))
-  (is (= true  (queen-attack/can-attack {:w [4 1] :b [6 3]}))))
+  (is (= true  (queen-attack/can-attack {:w [4 1] :b [6 3]})))
+  (is (= true  (queen-attack/can-attack {:w [2 3] :b [5 0]})))
+  (is (= true  (queen-attack/can-attack {:w [2 3] :b [0 5]}))))


fernandrone:
For further information, I noticed this issue when working on the problem, here's my solution: https://exercism.io/tracks/clojure/exercises/queen-attack/solutions/fdc2078be9d94045b92c18b1c9391e8e

There I use a absolute value function `abs` to verify, and it seems to cover all diagonal attack scenarios:

```clojure
(defn- diag? [[x1 y1] [x2 y2]]
  (= (abs (- y2 y1)) (abs (- x2 x1))))
```


I noticed that a different solution (which I liked very much by the way) didn't use it and I imagined it would probably fail if one of the differences were negative, but the other wasn't, which turned out to be the case:

https://exercism.io/tracks/clojure/exercises/queen-attack/solutions/3e6c1b0aeee14581af755a1ef0ae03fb

```
    (= (- x1 y1) (- x2 y2)) true
```

fernandrone:
I saw that tests were failing, decided to tackle this in a different pull request https://github.com/exercism/clojure/pull/370

porkostomus (approved):
Seems reasonable to me. Your contribution is much appreciated :)

Issue #374, opened 2021-06-28
Concept Exercises and Practice Exercises use different project files


ErikSchierboom:
The Practice Exercises all have a `project.clj` file, whereas the Concept Exercises instead have `deps.edn` file. I have no idea which is better, but when testing on exercism.lol, I found that Concept Exercises were not tested correctly due to the test runner expecting a `project.clj` file. It is probably best to use just one format.

porkostomus:
I had decided it would be preferable to switch to `deps.edn` because it ships with Clojure itself, making no need to use Leiningen. My only hesitation (besides needing to switch them all over) is that Leiningen seems to be refusing to die, loads of popular projects still use it, and people are used to being able to type `lein test` to run all the tests, while with `deps.edn` the easiest way is to set an alias to pull in the [Cognitect test runner](https://github.com/cognitect-labs/test-runner), although every Clojure editor plug-in includes a way of easily running tests.

It's also very significant that Leiningen (`project.clj`) uses twice the resources to run Clojure programs, since it starts up a whole extra JVM for the REPL... for me that's enough reason to switch right there.

ErikSchierboom:
@porkostomus Ah great! Is is hard to port the old exercises to the new structure? We'd also need to updat the test runner. We _could_ temporarily disable concept exercises if you don't have the time to do this.

RoelofWobben:
Is it difficult to make the switch ?

ErikSchierboom:
@porkostomus I think pulling in the Cognitect test runner by default seems like a great idea.

cstby:
Looks like this issue can be closed out? Or is there still work to be done?

porkostomus:
Yeah, all exercises have both files now so we're done here :)

Issue #332, opened 2021-02-07
Migrate from Leiningen


porkostomus:
As we are preparing for v3 launch this feels like the right time to make the switch completely from Leiningen to the Clojure CLI and Cognitect's test runner.

The rationale for this change is spelled out very clearly in Exercism's guidelines for choosing a test framework - it comes with Clojure by default and has been the de-facto tool for quite some time.

Besides the obvious job of replacing each `project.clj` with `deps.edn`, the user-facing documentation will need to reflect how to run the tests with Cognitect's test runner. Sadly, I kinda miss just being able to type `lein test`, but the Clojure CLI rocks in every other way...

porkostomus:
Closing this, because the solution seems to be to support both. Add a `deps.edn`, leave `project.clj` and let the user run whichever tool they want. Problem solved!

PR #376, opened 2021-07-02
add deps.edn to all practice exercises

porkostomus:
This PR adds a `deps.edn` file to all practice exercises, each including a `:test` alias to invoke Cognitect's test runner to discover and run the exercise's unit tests without the need for Leiningen. `project.clj` files are retained for legacy, since nothing is stopping anyone from doing it as before.

See: #374 and #332


SleeplessByte (approved):
I don't know what this does, but sure!

Issue #294, opened 2020-08-10
Improve the tags/topics for exercises


cwithmichael:
It seems like most of the exercises have the `misc` tag slapped on them.
For example the `Raindrops` exercise has `misc`. But it might be more helpful for students to include tags that describe some of the concepts that might be used in solving the exercise. 
i.e. `conditionals`

It looks like the tags are referred to as `topics` in this JSON file :https://github.com/exercism/clojure/blob/master/config.json
I'm guessing `misc` is the default tag applied if `topics` is `null`.

I can submit a PR that adds more helpful tags soon. That is, if no one else picks up this issue 😅 . 
Great work btw!

PR #384, opened 2021-08-13
Improve tags/topics for exercises

Lisa-Stats:
Fixes: https://github.com/exercism/clojure/issues/294#issue-676417114

Added tags to topics header in config.json.


porkostomus (approved):
This looks great! Thank you for taking the time to do this.

PR #398, opened 2021-09-06
Update docs

porkostomus:
- Removed several resources that have become inactive and added several new ones that have emerged in the last few years.

- Updated instructions for installation and testing with the new Clojure CLI, retaining Leiningen instructions for those still using that. 

- New section for running Clojure on Windows


SleeplessByte (approved):
I don't know clojure, but approving so if you don't get actual reviews, it can still be merged.

ErikSchierboom (approved):

kotp (approved):

PR #401, opened 2021-09-07
Disable concepts

porkostomus:
It seems that enabling Concept exercises has broken people's progress (see https://github.com/exercism/exercism/discussions/5830). If that is the case, this should fix it.

### DIFF
--- a/config.json
+++ b/config.json
@@ -3,7 +3,7 @@
   "slug": "clojure",
   "active": true,
   "status": {
-    "concept_exercises": true,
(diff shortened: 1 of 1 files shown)


BethanyG (approved):
Looks good.

ee7 (commented):
Looks like a no-op commit: https://github.com/exercism/clojure/commit/c80cde5620a9904eab7d22bf58f8938420708d07

BethanyG:
Three-way race condition!

PR #402, opened 2021-09-07
Disable concept exercises

ee7:
Follow-up of #401, if you want it.


porkostomus (approved):
Thanks, that's what I meant to do :)

PR #408, opened 2021-09-08
Fix faulty unit test

porkostomus:
The tests were failing because there was an error in the test! The `expected-time` unit test was trying to call a number as a function.


ErikSchierboom (approved):
Sweet!

PR #412, opened 2021-09-08
Fix bird-watcher exercise

porkostomus:
Page still contains embarrassing TODO comment, and it shows the answer to the student. SPOILER ALERT!!

### DIFF
--- a/concepts/vectors/introduction.md
+++ b/concepts/vectors/introduction.md
@@ -1,15 +1,6 @@
 # Introduction
 
-TODO: the content below is copied from the exercise introduction and probably needs rewriting to a proper concept introduction
-
-A `vector` in Clojure is a sequential, indexed, immutable collection of zero or more values. This means that once a vector has been created, it cannot be modified. Functions for operating on vectors will return a new vector, while the original vector remains unchanged. The values in a vector may be of heterogenous types. Vectors can be defined as follows:
-
-```clojure
-(def empty [])
-(def single-value 5)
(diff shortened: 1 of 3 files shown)


concepts/vectors/introduction.md
```diff
@@ -1,15 +1,6 @@
...
-TODO: the content below is copied from the exercise introduction and probably needs rewriting to a proper concept introduction
-
-A `vector` in Clojure is a sequential, indexed, immutable collection of zero or more values. This means that once a vector has been created, it cannot be modified. Functions for operating on vectors will return a new vector, while the original vector remains unchanged. The values in a vector may be of heterogenous types. Vectors can be defined as follows:
-
-```clojure
-(def empty [])
-(def single-value 5)
-(def single-value-alternative (vector 5))
-(def three-values [a b c]
-```
```
ErikSchierboom, line 12: For concepts that introduce new syntax (like vectors here), we require a syntax example to be included in the introduction. The introduction does not have to be exhaustive though, so maybe include something like:

```clojure
(def empty [])
(def three-values [1 "a" "b"]
```

ErikSchierboom (approved):
Approved with a small comment.

concepts/vectors/introduction.md
porkostomus: That seems sensible :)

ErikSchierboom:
The Clojure track is shaping up great!

PR #417, opened 2021-09-09
Fix test directory name

porkostomus:
This exercise is currently broken due to a file name mismatch which this corrects.


porkostomus:
Hmm now configlet can't find files which as far as I can see, do exist. I must be missing something silly here

verdammelt (approved):

ErikSchierboom:
Note: we'll update configlet soon-ish to detect these things.

PR #426, opened 2021-09-15
Create testing workflow

porkostomus:


porkostomus:
So now we have a `test.yml` workflow that fetches this repo and the `clojure-test-runner` repo, builds the test-runner Docker image and runs its tests. But that's not quite what we want.

We also have a `test.clj` file in _this_ repo that tests all the exercises with the examplar solutions. But that's also not quite what we want.

What we want is to wire these together so that the test runner part will test the exercises in the main repo.

porkostomus:
`test.clj` (a babashka script) now takes 2 arguments:

* the path to the root exercises directory
* the path to the `test-runner.clj` script (from the `clojure-test-runner` repo)

porkostomus:
And it works! The proof is that the new workflow didn't pass until I fixed some errors in an exercise.

ErikSchierboom (approved):
Neat!

PR #427, opened 2021-09-15
Matching exercise description.

Brunosmag:
Hello,
Was running the exercises and found that this step comment was different from the expected.

image


porkostomus (approved):
Nice catch, thank you :)

PR #429, opened 2021-09-18
Fix typos in spec

adamzerner:
A few dashes were missing in a test name, causing the specs to error out. This PR fixes that.


porkostomus (approved):
Excellent, thank you :)

PR #431, opened 2021-09-18
Fix linting errors

porkostomus:
* Had too many exercises practicing `numbers` and `strings`. Narrowed it down to ones that I thought were "definitely exercises in `x`".
* Added `international-calling-connoisseur` and `squeaky-clean` to `config.json` with a `status` of `wip`, since their directories and WIP contents exist.


verdammelt (approved):
Looks good to me and this reminds me i have to do the same on common-lisp :(

Issue #428, opened 2021-09-17
Interest-is-interesting: Should a negative balance update move closer to 0, or further away?


snits:
With the way banks normally work I'd expect a negative balance to go further negative instead of approaching 0:

(is (= -0.11904801M (interest-is-interesting/annual-balance-update -0.123M)))

I think this is a result of a multiplying a negative balance with a negative interest rate, resulting in a positive yield for the adjustment.
Instead I think the more accurate thing would be to multiply the interest rate against the absolute value of the balance, so when it is negative the yield will be negative as well.

porkostomus:
Thanks for bringing this to my attention. I made some critical errors when porting this, and now I know why! I've linked a PR to fix them.

Issue #430, opened 2021-09-18
Interest is interesting - dependence on implementation details


m1kal:
There are two problems with this exercise, but I think they can be fixed under a single issue. In my opinion both are related to unnecessary exposure of implementation details.

First, the stub contains `annual-yield`, which is not covered by any test case. Stub should contain only functions that are tested, other (helper) functions are implementation details.

Second, the test:

    (deftest annual-balance-update-large-positive-balance-test
      (is (= 1016.2101016209999M (interest-is-interesting/annual-balance-update 1000.0001M))))

checks if the implementation made a particular numerical error. Both input and output are BigDecimals, so the correct answer could be 1016.210101621M as well. Although it is reasonable to assume most solutions make this mistake, other tests such as

    (deftest annual-balance-update-small-positive-balance-test
      (is (= 0.000001005M (interest-is-interesting/annual-balance-update 0.000001M))))

can fail due to rounding errors.

porkostomus:
Thank you for the feedback! I ported this exercise from [the C# version](https://github.com/exercism/csharp/blob/main/exercises/concept/interest-is-interesting/InterestIsInterestingTests.cs) which does indeed calculate `1016.210101621m`. I was confused why I was getting a different result, and never really got to the bottom of it.

m1kal:
Thanks!
The changes look good.

PR #432, opened 2021-09-18
Correct BigDecimal arithmetic

porkostomus:
Fixes: #428 and #430. When porting this exercise I noticed some arithmetic discrepancies and wrongly assumed they were due to minor differences in the way the JVM and CLR handle decimals. Turns out I was doing something wrong, and fortunately this was brought to my attention so I could fix it before this exercise confuses many more people :P

### DIFF
--- a/config.json
+++ b/config.json
@@ -129,6 +129,22 @@
           "lists", "vectors", "strings"
         ],
         "status": "beta"
+      },
+      {
+        "slug": "international-calling-connoisseur",
+        "name": "International Calling Connoisseur",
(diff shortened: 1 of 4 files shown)


verdammelt (approved):

nitincodery:
Why taking absolute of balance? Can you explain me the logic behind it? If you do: `-100 + (-100 x -5%)` on calculator it yeilds different answer, but what you did is: `-100 + (abs(-100) x -5%)`, Why?

PR #450, opened 2021-11-08
Update instructions.md for exercice sublist

adrientiburce:
I find it really difficult to understand what was asked in the exercice sublist.
It's not really clear what the function should return.

I improved the instructions.


porkostomus (approved):
Yes this is much better, thank you!

PR #452, opened 2021-11-12
docs(elyses): remove misleading instruction

adrech:
Part 3, "Swap the first two cards", can't be solved with destructuring alone, yet the instructions suggest otherwise.

If someone is up for the challenge, please try it, but even @porkostomus (author) used `vec` and `conj` :) 
https://exercism.org/tracks/clojure/exercises/elyses-destructured-enchantments/solutions/porkostomus


porkostomus (approved):
Thanks for taking the time to improve the docs.

Issue #459, opened 2022-01-11
Typo: "abouts"


joachimnielandt:
https://github.com/exercism/clojure/blob/f10180101a8182fd5ceb862d2bdf0fd5a67920d9/exercises/concept/tracks-on-tracks-on-tracks/.meta/config.json#L2

`abouts` should be `about`, I'm assuming. Not sure I should make a pull request for this, or an issue. Perhaps I'll learn by doing :).

porkostomus:
Good catch. A PR would be welcome, thanks :)

PR #460, opened 2022-01-12
Typo: `abouts` > `about`

joachimnielandt:
Perhaps the first contribution of many. And perhaps the only one that is only about one character :). 

Typo: `abouts` should be `about`. #459  

I'm assuming pull requests go to `main`. If not, I'd love to learn otherwise.


porkostomus (approved):
Thanks!

PR #461, opened 2022-01-14
Update tracks_on_tracks_on_tracks_test.clj

abionescu:
I think here should be thread-first instead of thread-last.


porkostomus (changes requested):
Good observation. It conforms to the [source file](https://github.com/exercism/clojure/blob/3bff77604d9914ad80f56fb0442468efebef405d/exercises/concept/tracks-on-tracks-on-tracks/src/tracks_on_tracks_on_tracks.clj#L8):

```
(defn add-language
  "Adds a language to the list."
  [lang lang-list]
  )
```

It should be changed in both places. Would you like to add another commit so it will match? Thanks :)

porkostomus (approved):

porkostomus:
Awesome, thanks!

PR #462, opened 2022-02-07
reverse the order of args in add-language function

thepurplingpoet:
With the default order, the "conj" function wouldn't behave well. Also, for consistency, as all other functions have "lang-list" as first argument.

### DIFF
--- a/exercises/concept/tracks-on-tracks-on-tracks/src/tracks_on_tracks_on_tracks.clj
+++ b/exercises/concept/tracks-on-tracks-on-tracks/src/tracks_on_tracks_on_tracks.clj
@@ -7,7 +7,7 @@
 
 (defn add-language
   "Adds a language to the list."
-  [lang lang-list]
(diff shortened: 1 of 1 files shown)


porkostomus:
Ah, good thinking. In that case I believe we would also need to change the unit test which threads the list first:

```
(deftest list-add-test
  (is (= '("JavaScript" "Java" "Lisp" "Clojure")
         (-> (tracks-on-tracks-on-tracks/new-list)
              (tracks-on-tracks-on-tracks/add-language "Clojure")
              (tracks-on-tracks-on-tracks/add-language "Lisp")
              (tracks-on-tracks-on-tracks/add-language "Java")
              (tracks-on-tracks-on-tracks/add-language "JavaScript")))))
```

porkostomus:
Wait nevermind, I just tried it and realized I was confused. This is probably good to merge as is.

porkostomus (approved):

PR #471, opened 2022-04-19
Clojure/change: Add new test for invalid amount/coins combination

0xRand0m:
Hi, I think this would be a worthwhile addition to the tests because it requires more elaborate checks inside the solution algorithm.

### DIFF
--- a/exercises/practice/change/.meta/src/example.clj
+++ b/exercises/practice/change/.meta/src/example.clj
@@ -1,14 +1,24 @@
 (ns change)
 
-(defn issue [sum coins]
-  (when (or (neg? sum) (and (pos? sum) (every? #(< sum %) coins)))
-    (throw (IllegalArgumentException. "cannot change")))
-  (let [coins (sort coins)
-        all-amounts (reduce (fn [cached-amounts amount]
-                              (->> coins
-                                   (filter #(<= % amount))
-                                   (map #(conj (cached-amounts (- amount %) []) %))
-                                   (apply min-key count)
-                                   (assoc cached-amounts amount)))
-                            {} (range 1 (inc sum)))]
-    (all-amounts sum)))
+(def algo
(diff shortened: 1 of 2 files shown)


BTowersCoding:
Awesome, love it! Except that it seems we might need to update the example solution because CI is failing. Unless something else is broken... I'll try to look into this today.

0xRand0m:
This is my solution which passes the test. Use any part you like ;)
```clojure
(ns change)

(def algo
  (memoize (fn [amount coins]
             (let [smaller (filter #(<= % amount) coins)]
               ; use a list so that amount always stays the first element
               (if (empty? smaller) `(~amount)
                 (apply min-key
                        count
                        ; check if valid solution, i.e. the final amount was zero
                        (filter (comp zero? first)
                                (map
                                  (fn [coin]
                                    (concat
                                      (algo (rem amount coin) smaller)
                                      (repeat (quot amount coin) coin)))
                                  smaller))))))))


(defn issue [amount coins] 
  (try
    (let [[x & xs] (algo amount coins)]
      (if (zero? x) xs
        (throw (IllegalArgumentException. "cannot change"))))
    ; thrown by apply min-key if we don't have the right coins to issue change
    (catch clojure.lang.ArityException e (throw (IllegalArgumentException. "cannot change")))))
```

BTowersCoding:
Yes, that works great!

If you'd like, the easiest way would probably be for you to make another commit to https://github.com/exercism/clojure/blob/main/exercises/practice/change/.meta/src/example.clj replacing the existing code with your solution above.

BTowersCoding:
I thought I smelled something funny, there seems to be a bug in the test runner as well

BTowersCoding:
I might have found the issue: https://github.com/exercism/clojure/pull/473

BTowersCoding (approved):

PR #473, opened 2022-04-21
remove dev workflow

BTowersCoding:
CI has been failing since features were added to the test runner and I believe it was related to a conditional development workflow in `test.clj` that doesn't belong there anyway. It was remaining from when I developed the CI script and it was actually preventing a proper integration test with the `clojure-test-runner` repo as was intended. This change simply hardcodes the `root` and `test-runner-dir` locations to where they are in prod.

### DIFF
--- a/test.clj
+++ b/test.clj
@@ -6,20 +6,9 @@
  '[clojure.java.shell :as shell]
  '[clojure.java.io :as io])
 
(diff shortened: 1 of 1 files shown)


kotp (approved):
Approving without testing it locally.  It appears reasonable to me.

PR #475, opened 2022-04-27
Temporary patch to get CI working again

BTowersCoding:
Upgraded the babashka Docker action to the one currently maintained by the community.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -24,11 +24,14 @@ jobs:
           repository: exercism/clojure-test-runner
           path: clojure-test-runner
 
-      - name: babashka script
-        uses: tzafrirben/babashka-docker-action@v1.1
-        id: bb_script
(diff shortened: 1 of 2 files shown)


BTowersCoding:
Hmm so it seems that babashka is setup and runs in the container correctly - then fails in the script itself.

BTowersCoding:
I've tested the `test.clj` script locally and it works:

`{:tested 86, :fails ()}`

BTowersCoding:
Finally figured it out. The `libs.jar` file is not being added to the classpath in CI because it's in a different location at that point.

Trying to think of the cleanest way to fix it. The quickest way is not the ideal way, but I'd like to see this finally go green and bring this epic debug session to a close.

It means temporarily using a local copy of the test-runner script instead of cloning the _actual_ test-runner repo.

SleeplessByte (approved):
Probably fine?

Issue #485, opened 2022-05-25
CI is passing even though an exemplar solution is failing


BTowersCoding:
See #484

PR #486, opened 2022-05-25
Fix CI bug

BTowersCoding:
Issue: #485 

There was a small but nasty bug in the test script that was causing some failing exercises to go unreported.


test-runner.clj
```diff
@@ -18,139 +18,92 @@
...
+                           {:name test :status "pass" :test_code (str (test (test-code-map zloc)))}
+                           (contains? (set (map :name @fails)) test)
+                           {:name test :status "fail" :test_code (str (test (test-code-map zloc)))
+                            :message (:message (first (filter #(= test (:name %)) @fails)))}
+                           (contains? (set (set @errors)) test)
+                           {:name test :status "error" :test_code (str (test (test-code-map zloc)))
+                            :message (:message (first (filter #(= test (:name %)) @errors)))})))}
+          {:pretty true}))
 
 (System/exit 0)
```
kotp, line 109: It might be nice to fix this missing EOL since we are here (a separate commit, perhaps, in case the rest of the patch is not accepted, this portion of the patch could be easily used).

exercises/concept/tracks-on-tracks-on-tracks/.meta/exemplar.clj
```diff
@@ -19,12 +19,11 @@
...
-                                (add-language "Lisp"
-                                              (add-language "Clojure"
-                                                            (new-list))))))))
+  (-> (new-list)
+      (add-language "Clojure")
+      (add-language "Lisp")
+      remove-language
+      (add-language "Java")
+      (add-language "JavaScript")
+      count-languages))
```
kotp, line 29: Same here for missing EOL.

(When I write up documents and concatenate the files together, I do not want to have to check if the last line is really a line with all the features of a line, or if it is in reality extraneous characters after the last line of the file that has that EOL character.)

kotp (approved):

PR #489, opened 2022-06-01
Add Coordinate Transformation exercise as WIP

BTowersCoding:
Adds exercise to `config.json` with `status` set to "wip" to make it available for testing.


exercises/concept/coordinate-transformation/.docs/introduction.md
```diff
@@ -0,0 +1,28 @@
+# Introduction
+
+**Closures** are a programming pattern in Clojure which allows variables from an outer [lexical scope][wiki-lexical-scope] to be used inside of a function. Clojure supports closures transparently, and they are often used without knowing what they are.
+
+```clojure
+;; Top-level definitions are global-scope
+(def dozen 12)
+
+;; Functions create a new scope.
+;; Referencing the outer variable here is a closure.
```
MatthijsBlom, line 10: What exactly is a closure here?

- The very act of referencing elements from outer scopes, or
- a single reference to, i.e. mention of, an element from an outer scope, or
- a function that contains references to elements of outer scopes?

I suspect the latter, from experience with other languages. Clojure terminology might differ though.

exercises/concept/coordinate-transformation/.docs/introduction.md
```diff
@@ -0,0 +1,28 @@
...
+
+```clojure
+
+;; This function closure increments the counter's state in the outer lexical context.
+;; This way the counter can be shared between many calling contexts.
+
+(def increment
+  (let [counter (atom 0)]
+    (fn [] (swap! counter inc))))
+```
```
MatthijsBlom, line 26: I could use an example of `increment` usage. Without one, I am having some trouble making sense of this code snippet.

exercises/concept/coordinate-transformation/.docs/introduction.md
BTowersCoding: This was translated verbatim from the JavaScript track: https://github.com/exercism/javascript/blob/main/exercises/concept/coordinate-transformation/.docs/introduction.md

exercises/concept/coordinate-transformation/.docs/introduction.md
BTowersCoding: Example usage:
```
(increment)
;;=> 1
(increment)
;;=> 2
```

exercises/concept/coordinate-transformation/.docs/introduction.md
MatthijsBlom: I see. In that case I disagree with the wording at the JavaScript track.

This is the kind of ambiguity that trips _me_ up, but I may well be atypical.

config.json
```diff
@@ -147,9 +147,9 @@
         "status": "wip"
       },
       {
-        "slug": "date-parser",
-        "name": "Date Parser",
-        "uuid": "37ba33f8-5b19-4cdd-9b1c-ca1a5c15f096",
```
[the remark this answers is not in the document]
BTowersCoding: Ouch, no that must have happened during merge conflict resolution. Thanks!

exercises/concept/coordinate-transformation/.docs/introduction.md
BTowersCoding: To be honest I find it confusing as well! Perhaps I can meditate upon this and come up with a better explanation :)

config.json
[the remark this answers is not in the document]
BTowersCoding: fixed

exercises/concept/coordinate-transformation/.docs/introduction.md
MatthijsBlom: Some meditation input:

- The below `increment` is a _closure over_ the outer names `atom`, `0`, `swap!`, `inc`, (possibly also `let` and `fn`, depending on how Clojure works), but **not** over `counter`.
- The inner `fn` is a closure over all of `swap!`, `counter`, and `inc`.
- Closures aren't special. Most by far of all the functions you will encounter 'in daily life' are closures. (I mention this because when learning about closures for the first time I encountered a lot of unhelpful mysticism similar to what surrounds monads even today.)
- [wiki.haskell.org/Closure](https://wiki.haskell.org/Closure) (note: that `f` also closes over `+`)
- It is not closures per se but rather mutability of variables that makes 'closures' hard to understand / interesting. The 'environments' that are often talked about are only relevant because of either mutability or interest in implementation details.

exercises/concept/coordinate-transformation/.docs/introduction.md
MatthijsBlom: That is neat. I would welcome adding this to (somewhere near) the code sample.

BTowersCoding: added it to the doc.

ErikSchierboom (approved):

PR #490, opened 2022-06-07
Improves `nth-prime` scaffold

matuella:
Not sure if this one specifically wants the missing arg to be included by the user, but until this exercise, all had the args filled, so I think maintaining consistency should be desired here.


BTowersCoding:
Nice, thanks for the contribution 💯

BTowersCoding (approved):

PR #491, opened 2022-06-07
test: fix list-add test

yosevu:
Use thread-last in list-add test to match Task 2
example:

Task 2 Add a new language to the list example:
```clojure
(add-language "JavaScript" '("Clojurescript"))
;; => '("JavaScript" "Clojurescript")
```

Fails Test 2 list-add-test
Screen Shot 2022-06-08 at 6 51 34 AM

Thread-first would work if the parameters were switched. What is the intended argument order?

```clojure

;; Current test depends on this parameter order
(defn add-language [lang-list lang]
  (conj lang-list lang))

;; Example gives this parameter order
(defn add-language [lang lang-list]
  (conj lang-list lang))
```

### DIFF
--- a/exercises/concept/tracks-on-tracks-on-tracks/.docs/instructions.md
+++ b/exercises/concept/tracks-on-tracks-on-tracks/.docs/instructions.md
@@ -18,7 +18,7 @@ Before you can add languages, you'll need to start by creating a new list. Defin
 As you explore Exercism and find languages you want to learn, you'll need to be able to add them to your list. Define a function to add a new language the beginning of your list.
 
 ```clojure
-(add-language "JavaScript" '("Clojurescript"))
+(add-language '("Clojurescript") "JavaScript")
 ;; => '("JavaScript" "Clojurescript")
 ```
 


BTowersCoding:
I think we realized that the `add-language` function should behave like `conj`, with the `lang-list` first.

yosevu:
I see. In that case, can I update the example in the Task 2 description to the following?

```clojure
(add-language '("Clojurescript") "JavaScript")
;; => '("JavaScript" "Clojurescript")
```

Out of curiosity, what is the reason for `add-language` to behave the like `conj`? Is it a best practice in Clojure? The other option also seems reasonable to me.

BTowersCoding:
This SO answer seems to explain it well: https://stackoverflow.com/questions/50275513/rules-of-thumb-for-function-arguments-ordering-in-clojure

The `cons` function seems to be an exception to the rule however, so I suppose an argument (no pun intended) could be made that way too. I think the original author of the exercise intended it that way but that was confusing people so at some point it was changed but that resulted in the inconsistency above.

yosevu:
Ah, got it. That SO answer was helpful. I reverted the test change and updated the argument order in the instructions instead.

BTowersCoding (approved):

PR #493, opened 2022-06-08
Improves `acronym` scaffold

matuella:
Same rationale as #490

### DIFF
--- a/exercises/practice/acronym/src/acronym.clj
+++ b/exercises/practice/acronym/src/acronym.clj
@@ -1,5 +1,7 @@
 (ns acronym)
 
-(defn acronym [] ;; <- arglist goes here
+(defn acronym
+  "Converts phrase to its acronym."
+  [phrase] ;; <- arglist goes here
   ;; your code goes here
 )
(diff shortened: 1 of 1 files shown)


exercises/practice/acronym/src/acronym.clj
```diff
@@ -1,5 +1,7 @@
 (ns acronym)
 
-(defn acronym [] ;; <- arglist goes here
+(defn acronym
```
BTowersCoding, line 3: As long as we're here, it's probably best to remove these comments because at this point they're just adding noise.
```suggestion
(defn acronym
  "Converts phrase to its acronym."
  [phrase]
  )
```

BTowersCoding (changes requested):

matuella:
Agreed.

I will try to aim for better docs and more standardized "scaffolds" if I eventually find more things like these.

BTowersCoding:
It looks like a few lines got repeated:

```
(defn acronym
  "Converts phrase to its acronym."
  [phrase]
  )
  "Converts phrase to its acronym."
  [phrase] ;; <- arglist goes here
  ;; your code goes here
)
```

matuella:
Sorry, didn't actually see that the suggestion wasn't to be properly committed!

BTowersCoding (approved):

PR #494, opened 2022-06-13
Update introduction.md

dsletten:
Added mention of `update`. This is needed for the bird watcher exercise.

### DIFF
--- a/concepts/vectors/introduction.md
+++ b/concepts/vectors/introduction.md
@@ -34,3 +34,10 @@ To remember the updated value, we need to pass it along or capture it in a var:
 (get updated-numbers 2)
 ;;=> 9
 ```
+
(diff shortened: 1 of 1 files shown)


tasxatzial:
The `update` isn't really needed since the exercise already mentions that vectors can be updated using `assoc`. If we decide to mention `update`, it would probably better to rewrite the article so that both `assoc` and `update` are mentioned in the same section.

@michalporeba Thoughts? I'm thinking that this can be safely closed.

michalporeba:
@dsletten are you still interested in this change? We had some delays in the track, my apologies for it. But me and @tasxatzial are now trying to get on top of the things, so if you are still interested we can talk about the change. If not, we will close the PR for now as we are thinking about how to improve the track.

PR #495, opened 2022-06-13
Update instructions.md

dsletten:
Added note about parameter order. The Clojure tests expect a strange order which differs from, say the corresponding Common Lisp and JavaScript exercises.


BTowersCoding (approved):
Cool, thanks for the contribution :)

PR #496, opened 2022-06-18
Update instructions.md

dsletten:
The inclusion of the word "positive" is redundant given the context. Moreover, it is inconsistent with the test suite. The test suite expects a result of 0.5 for a balance of 0, which is not a positive number.


exercises/concept/interest-is-interesting/.docs/instructions.md
```diff
@@ -3,9 +3,9 @@
 In this exercise you'll be working with savings accounts. Each year, the balance of your savings account is updated based on its interest rate. The interest rate your bank gives you depends on the amount of money in your account (its balance):
 
 - -3.213% for a negative balance.
-- 0.5% for a positive balance less than `1000` dollars.
-- 1.621% for a positive balance greater or equal than `1000` dollars and less than `5000` dollars.
-- 2.475% for a positive balance greater or equal than `5000` dollars.
+- 0.5% for a balance less than `1000` dollars.
+- 1.621% for a balance greater or equal than `1000` dollars and less than `5000` dollars.
```
BTowersCoding, line 7: Thanks! While taking a closer look I happened to notice that the wording here is strange, maybe we could fix this now as well?
```suggestion
- 1.621% for a balance greater than or equal to `1000` dollars and less than `5000` dollars.
```

exercises/concept/interest-is-interesting/.docs/instructions.md
```diff
@@ -3,9 +3,9 @@
...
+- 2.475% for a balance greater or equal than `5000` dollars.
```
BTowersCoding, line 8: Also here:
```suggestion
- 2.475% for a balance greater than or equal to `5000` dollars.
```

bobbicodes (dismissed):

exercises/concept/interest-is-interesting/.docs/instructions.md
dsletten: I missed that. Your suggestion reads more naturally.

tasxatzial:
This PR is stale at this point and conflicts with a very recent commit that already applied Bobbi's suggestions. I'm going to resolve the conflicts and then merge, provided no further changes are necessary.

tasxatzial (approved):

PR #500, opened 2022-07-07
Interest is Interesting exemplar: Use Java's built-in Math.abs instead of if block

miridius:
It's not a good idea to roll your own absolute value logic when the standard library has a method for that already

### DIFF
--- a/exercises/concept/interest-is-interesting/.meta/exemplar.clj
+++ b/exercises/concept/interest-is-interesting/.meta/exemplar.clj
@@ -10,7 +10,7 @@
 (defn- annual-yield [balance]
   (let [multiplier (bigdec (/ (interest-rate balance)
                               100))]
-    (* (if (neg? balance) (- balance) balance) multiplier)))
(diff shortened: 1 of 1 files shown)


BTowersCoding (approved):

bobbicodes:
Sorry I took so long to get to this... wow, it's been a whole year?

It looks like I need to update babashka because it's complaining that `Math/abs` isn't available. But I just tested it in my bb repl v1.3.181 and it works, so imagine it must have been added.

tasxatzial:
The problem with using `Math/abs` is that does not accept bigdecimals. It only works with ints, floats, longs. Based on the proposed diff, I expect the tests to fail.

bobbicodes:
@tasxatzial ah, you're right, thanks! I would have been very confused when this was still failing after updating the test runner.

tasxatzial:
@bobbicodes I've found a simpler solution. May i submit a new PR?

bobbicodes:
@tasxatzial sure!

PR #503, opened 2022-07-31
Update pig_latin_test.clj

swaraj:
Added 2 more tests to cover the rule 4 from exercise as I could get away with without implementing the rule 4 and pass all tests.


BTowersCoding (approved):
> "ythmrhay"

Love it! This PR does have a good ythmrhay to it.

PR #504, opened 2022-08-08
Update elyses_destructured_enchantments.clj

dsletten:
The test suite specifically does not test that the function returns a vector or that the remaining cards are held in a vector. This comment adds pointless complexity to the exercise which is trivially solved using destructuring.


BTowersCoding (approved):
That makes sense, thank you :)

PR #505, opened 2022-08-08
Update exemplar.clj

dsletten:
The exemplar solution should not guide students to shadow functions in clojure.core, e.g., first, second, rest. It should also illustrate more straightforward destructuring on function parameters. There is no need for separate let bindings. Finally, the comment in discard-top-card is inconsistent with the test suite. Furthermore, the function does not satisfy the results stated in the doc comment. Namely, the remaining cards are not returned as a vector.


BTowersCoding (approved):
Great improvements, thanks :)

Issue #519, opened 2022-11-19
Elyses Destructured Enchantments: Tests mix up Expected and Actual results


SebAlbert:
The tests in Elyses Destructured Enchantments have the wrong argument order in their `(is (= ...))` forms, so that the actual result of the learner's code will be called the Expected result in the output of a failing test, and vice versa.

BTowersCoding:
@SebAlbert thanks for pointing this out! To be honest, I didn't even realize that the order here is what determines the expected/result reporting.

PR #522, opened 2022-11-27
Fix assertion argument order in Elyses Destructured Enchantments tests

BTowersCoding:
Fixes: #519


kotp (approved):
It might be informative to define what "expected order" is.

Expected, Actual being that order.

The title or the commit message itself could explain it for those that may not be aware.

PR #552, opened 2023-05-23
List ops exercise

bobbicodes:
Creating an exercise generator.

It outputs a test suite, which is mostly correct except:

* anonymous functions are not translated
For example:

```clojure
(list-ops/filter [] "(x) -> x modulo 2 == 1")
````
should be
```clojure
(list-ops/filter [] (fn [x] (= (mod x 2) 1))
```
EDIT: test suite should now be correct.


kahgoh:
Hi @bobbicodes, are you still working on this PR? If not, I was wondering if I could continue this PR to add the list-ops exercise to the Clojure track? (although probably without the generator bit as I'm unfamiliar with it)

bobbicodes:
@kahgoh Sure! I've been working on other projects these days and forgot all about this. It might even be nearly finished, I was porting it as part of last year's Functional February and at the last minute decided to do a different exercise.

PR #559, opened 2023-06-11
Update cars_assemble_test.clj

pdmoore:
One of the test names misspelled "speed". Fix to match all other test names.


bobbicodes (approved):
Thanks :)

PR #563, opened 2023-06-22
cars-assemble: Update introduction.md

DevPika:
`<>` inequality operator doesn't seem to exist in Clojure, replaced it with `not=`

### DIFF
--- a/exercises/concept/cars-assemble/.docs/introduction.md
+++ b/exercises/concept/cars-assemble/.docs/introduction.md
@@ -16,3 +16,14 @@ In this exercise you must conditionally execute logic. A common way to do this i
       (> x 7) "Expression to evaluate when x is greater than 7"
       :else   "Expression to evaluate in all other cases")
 ```
+
+## Note
+The `==` operator might be preferable to `=` in some cases where numbers need to be compared irrespective of the exact type. For instance,
+```clojure
+(== 5.0 5) ;; true as both numbers are equal when type is ignored
+(= 5.0 5) ;; false as the types of numbers are also taken into account here i.e. float is different from int
+```
+The [Clojure guide on Equality][guide-equality], specifically the section on [Numbers][guide-numbers] goes into more details.
+
+[guide-equality]: https://clojure.org/guides/equality
+[guide-numbers]: https://clojure.org/guides/equality#numbers
--- a/exercises/concept/cars-assemble/.docs/introduction.md
+++ b/exercises/concept/cars-assemble/.docs/introduction.md
@@ -7,7 +7,7 @@ Numbers in Clojure include:
 
 Two common numeric types are `int` and `float`. An `int` is a 32-bit integer and a `float` is a 64-bit floating-point number.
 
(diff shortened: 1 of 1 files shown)


DevPika:
I understood the use of the plural form "operators" in the sentence to be for the "standard numeric comparison operators" (which is probably `>, <, <=, >=`), equality `=` and non-equality ("inequality" here). The issue I have with the wording is that some languages like F# have [`<>` as the not equal operator](https://learn.microsoft.com/en-us/dotnet/fsharp/language-reference/symbol-and-operator-reference/arithmetic-operators#summary-of-binary-comparison-operators), which is not the case for Clojure, even though it might seem that way from the introduction here. Further discussion on differences in Clojure between `=` and `==` operators can be found on [the Clojure equality guide page](https://clojure.org/guides/equality).

DevPika:
I have tried to make the sentence clearer now, but I am unsure if the difference between `==` and `=` as discussed in [the Clojure guide](https://clojure.org/guides/equality#:~:text=%3D%20does%20not%20always%20return%20true%20when%20two%20numbers%20have%20the%20same%20numeric%20value.) can go in as a note here. For example, consider this:
```clojure

;; false
(= 2 2.0)

;; true, because == is type independent and specifically deals with numbers
(== 2 2.0)
```

bobbicodes:
> . . . "standard numeric comparison operators" . . . The issue I have with the wording is that some languages like F# have . . .

I copied that wording from the [C# track](https://github.com/exercism/csharp/blob/main/exercises/concept/cars-assemble/.docs/introduction.md) but now I don't like it either. I think it would be good to avoid saying "standard" anything and enumerate them specifically.

tasxatzial:
> I have tried to make the sentence clearer now, but I am unsure if the difference between `==` and `=` as discussed in [the Clojure guide](https://clojure.org/guides/equality#:~:text=%3D%20does%20not%20always%20return%20true%20when%20two%20numbers%20have%20the%20same%20numeric%20value.) can go in as a note here. For example, consider this:
> 
> ```clojure
> ;; false
> (= 2 2.0)
> 
> ;; true, because == is type independent and specifically deals with numbers
> (== 2 2.0)
> ```

Good point. This example shows why sometimes `==` might be preferable. It needs more explanation though because it's not perfectly clear why the first example returns false and why the second one returns true. The linked [Clojure guide](https://clojure.org/guides/equality#numbers) has an extensive section talking about their differences. I'm not sure what's the best approach at this point. Maybe write a brief note and link to the guide? @bobbicodes What's your take on this?

bobbicodes:
I think a note about `==` here seems totally appropriate. For comparison, I just looked at the [Python introduction](https://github.com/exercism/python/blob/main/exercises/concept/currency-exchange/.docs/introduction.md) to the exercise teaching numbers and it leads me to think we should lean in the direction of providing more information.

DevPika:
Thanks for the feedback! We are taking this PR _closer_ to _closure!_ :grin:  I have added in a note at the bottom to highlight the number equality behavior.

bobbicodes (approved):
This is great, thank you!

PR #565, opened 2023-07-06
Interest is interesting: Update exemplar solution to use built-in function abs from Clojure 1.11

tasxatzial:
Clojure 1.11 added an `abs` function that we can now use instead of defining our own function or `Math/abs`.

I've also replaced the TODO in docstrings. This is just a copy-paste from the problem description. Feel free to edit them if needed.


bobbicodes (approved):
Thanks! This is great

PR #566, opened 2023-07-07
Squeaky clean: Use different control chars

tasxatzial:
The string "my\0Id" works fine in Java, but isn't parsed at all in Clojure. I've replaced it with a proper unicode sequence.

Another thing i've noticed is that some people use something like `#"\p{Cntrl}"` as a regex that cleans the control chars. This regex looks fine, but it's not. It covers the first range `u0000-\u001F` and only the first char from the second range `\u007F-\u009F`. Obviously those people didn't even consider the possibility of that regex not covering everything.

Should we replace the control char in tests with a different one? If so, i'll do it in this PR.

### DIFF
--- a/exercises/concept/squeaky-clean/.docs/instructions.md
+++ b/exercises/concept/squeaky-clean/.docs/instructions.md
@@ -32,7 +32,7 @@ its code is in the range '\u0000' through '\u001F'
 or in the range '\u007F' through '\u009F'.
 
 ```clojure
-(clean "my\0Id")
+(clean "my\u007FId")
 ;;=> "myCTRLId"
 ```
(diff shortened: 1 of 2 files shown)


bobbicodes:
Yes it would be good to have a test that fails with the regex solution.

tasxatzial:
I think i'm done. I'm guessing these chars aren't going to be rendered due to the issue discussed [here](https://forum.exercism.org/t/squeky-clean-might-have-a-poorly-defined-test-case/6136)

so i'll probably update the thread too.

bobbicodes (approved):
Great! Thanks for doing this :)

Issue #520, opened 2022-11-25
Complex Numbers: Tests mix up Expected and Actual results, and unclear about requiring floats rather than rationals


SebAlbert:
The tests for the Complex Numbers exercise have the Expected and Actual results in the wrong order in their `(is)` forms.
Moreover, it could be made more explicit that floats are expected when doing division rather than rational numbers (especially if the test suite output suggests that it would be the other way around).

BTowersCoding:
Thanks again, I can't take the blame for this one because I didn't write it :)

This makes me wonder how many other test suites have the expected/actual args reversed. This must be a pretty common mistake.

PR #567, opened 2023-07-11
Complex numbers: Sync tests with problem-specifications

tasxatzial:
Partially addresses #520 

Tests are now in sync, missing tests have been implemented, order of the actual and expected results has been fixed.

I'll try to address the issue with the floats in a subsequent PR since this one is changes a lot of things.

### DIFF
--- a/exercises/practice/complex-numbers/.docs/instructions.md
+++ b/exercises/practice/complex-numbers/.docs/instructions.md
@@ -22,8 +22,8 @@ Dividing a complex number `a + i * b` by another `c + i * d` gives:
 Raising e to a complex exponent can be expressed as `e^(a + i * b) = e^a * e^(i * b)`, the last term of which is given by Euler's formula `e^(i * b) = cos(b) + i * sin(b)`.
 
 Implement the following operations:
- - addition, subtraction, multiplication and division of two complex numbers,
- - conjugate, absolute value, exponent of a given complex number.
 
+- addition, subtraction, multiplication and division of two complex numbers,
+- conjugate, absolute value, exponent of a given complex number.
 
 Assume the programming language you are using does not have an implementation of complex numbers.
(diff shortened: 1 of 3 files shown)


bobbicodes:
For some reason it's not giving me the option to reopen this... weird

bobbicodes:
Oh, now it is :)

bobbicodes:
I still find it odd that the expected/actual results depend on the order in the `=` expression. I'm trying to see if that is documented anywhere but it doesn't seem to be.

tasxatzial:
Ok, so i had a suspicion that the order of actual expected is nothing more than a convention. The Clojure api [here](https://clojure.github.io/clojure/clojure.test-api.html) shows an example with expected being the initial form and actual being the computed result. Re-writing `(= 5 (+ 2 2))` as `(= (+ 2 2) 5)` simply switches the initial form, but this doesn't really make any difference because the actual form remains the same. I've tested this with the online editor and with lein and both don't care about the order.

The confusion about the convention arises when we realize that not every test tool behaves like this. Cursive, for example, reports something completely different.

`(is (= 5 (+ 2 2))) -> Expected 4, Actual 5`
`(is (= (+ 2 2) 5)) -> Expected 5, Actual 4`

So not only it does not retain the initial form, but now the order makes a difference. It treats the first argument as the actual and the second as the expected. But apparently for most test tools the order comes from the Clojure api and is the exact opposite: First is expected, second is actual.

How do i know this? I just asked the Cursive author [here](https://groups.google.com/g/cursive/c/NQ1tX1edeUg). So I'll leave the order as is and hopefully some day he fixes Cursive.

tasxatzial:
This one is finished. Tested with random community solutions and i've found one that fails due to floating point inaccuracies. The good news is that the next PR, which is ready, will resolve those issues.

bobbicodes (approved):
Cool, thanks!

PR #568, opened 2023-07-12
Update difficulties based on completion rate

bobbicodes:
See the [track build status](https://exercism.org/tracks/clojure/build) page for source data

### DIFF
--- a/config.json
+++ b/config.json
@@ -179,7 +179,7 @@
         "uuid": "0e0fec77-651a-45a7-af99-a768ebebef05",
         "practices": ["numbers"],
         "prerequisites": ["numbers"],
-        "difficulty": 4,
+        "difficulty": 6,
         "topics": [
           "math"
         ]
@@ -201,7 +201,7 @@
         "uuid": "0f7b1858-15d0-467d-9f94-56408249c2f7",
         "practices": [],
         "prerequisites": ["strings"],
-        "difficulty": 3,
+        "difficulty": 1,
         "topics": [
           "strings", "core_functions"
         ]
@@ -212,7 +212,7 @@
         "uuid": "49f62bbc-0f60-4922-b5a6-f266b80442f4",
         "practices": [],
         "prerequisites": ["numbers"],
-        "difficulty": 3,
+        "difficulty": 1,
         "topics": [
           "algorithms, core_functions"
         ]
@@ -223,7 +223,7 @@
         "uuid": "b99ed7da-98a8-43f0-8162-05c5408f6ac5",
         "practices": ["strings"],
         "prerequisites": ["strings", "booleans"],
-        "difficulty": 3,
+        "difficulty": 1,
         "topics": [
           "strings", "regular_expressions"
         ]
@@ -234,7 +234,7 @@
         "uuid": "7b8c5f7e-3c7d-4e69-ba8c-f29bb492a628",
         "practices": ["numbers"],
         "prerequisites": ["numbers", "vectors"],
-        "difficulty": 3,
+        "difficulty": 5,
         "topics": [
           "math"
         ]
@@ -245,7 +245,7 @@
         "uuid": "11818959-7e42-4a1d-8dfd-288cea744d66",
         "practices": ["strings"],
         "prerequisites": ["strings", "vectors"],
-        "difficulty": 3,
+        "difficulty": 1,
         "topics": [
           "strings"
         ]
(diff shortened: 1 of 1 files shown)


tasxatzial:
Very weird results.

Reverse-string went from 1 to 3 even though it's a trivial one liner.
Hamming went from 2 to 4 even though it's one of the simplest exercises.
Complex numbers went from 2 to 4 even though the solutions look pretty much the same in every language.
Change which is pretty hard, remained at 3.
Say, wordy, dominoes, spiral matrix are now a 7. I agree with those.
Robot simulator is now a 7. That's quite odd. This is a simple exercise.

And most exercises remained unchanged, also not expected.

bobbicodes:
Sorry, I probably should have waited before submitting this, it's not really ready yet

tasxatzial:
Since i don't have access to completion rate, do you have any idea how it is defined? Also how did you come up with that formula? What do these numbers mean?

bobbicodes:
I'm probably not going to merge this, it's just a little experiment to help guide me on the track's development based on the only objective measure we have. Completion rate is the number of solutions submitted divided by the number of student who started it. they range from 93% (Leap, Word count, RNA Transcription) to 35% (Go counting). The formula scales them so they fall between 1 and 10.

The data is publicly available on the track build page linked above.

tasxatzial:
Hmm, i see. Let me know if you are interested in a second opinion. I believe i can come up with a different and possibly more accurate way of adjusting the difficulties. But it does require manual work.

bobbicodes:
The whole idea of accurately judging the difficulties is kind of a fallacy, and doesn't seem like a valuable use of time. Any possible strategy takes on too many assumptions because there's no canonical approach to any exercise, and no matter how carefully they are chosen, people will always find certain exercises out of place. I was just curious what it would look like if I set them according to an objective formula, but the results don't seem much better than they were before. I'm leaning towards just keeping them as they are.

tasxatzial:
Well, you are right. People will always find exercises out of place. That's why a difficulty level should never be adjusted based on what people think because this leaves no room for any adjustments. There's also a fundamental mathematical flaw with how the final difficulty is calculated.

I do have a lot of things to say about all this. But my approach is quite different and it doesn't seem to match any of the ideas i've seen so far. And i do believe that this is a solvable problem.

bobbicodes:
If you want to take a crack at it, I'm all ears.

tasxatzial:
I was kind of hoping you said no because we already know that setting difficulties isn't an easy task in any case. Before opening a new issue about how i think about all this, i'll briefly explain here why i think that completion rate isn't very reliable. You said that it is the only objective measure we have. Well, if that's the case then i'm afraid i have some bad news. Completion rate is anything but objective. Why? Because it depends on the skill level of the individual.

It's safe to assume that most people who attempt this track are not complete beginners because beginners don't typically start with Clojure. There are most likely people who have extensive programming experience. This means they'll tend to increase the completion rate, and that means decreasing the difficulty. So if there are any true beginners among the users, they'll complain that the exercises are too hard for them.

The existing difficulty level of the exercise also affects the  completion rate. If an exercise is marked as hard, people will see it's hard and most of them won't bother. This means the people who will attempt it are probably those who are experts, super skilled, and feel confident. So again, we should be expecting the completion rate to be higher than expected.

Another reason that affects the completion rate is the nature of the exercise. For example nth-prime is probably not going to be attempted by people who are not good at math. Those who will attempt it will most likely be able to solve it because it's a trivial problem if you are math inclined. So we should be expecting this exercise to have a high completion rate and in fact, this is already evident. It's marked as easy. But in general, this isn't an easy problem.

So as we see, all the above tells us that completion rate does carry info about the individual and therefore cannot be used as an objective measure.

bobbicodes:
I totally agree with everything you are saying. I closed the PR because this has already been discussed far longer than I would have liked. It's great that you want to help, but sometimes I make issues and PRs not because I'm looking for feedback, but as a way of documenting a thought so that it will be there later if and when it feels appropriate. In this case, it was motivated precisely because I thought it might be a way to *avoid* getting caught up in a discussion about difficulties. I hope this doesn't come across as harsh, I'm just not wanting to focus on this at this time.

tasxatzial:
I don't have any objections. If you've decided to merge this i would be ok with it even though i already knew as soon as i saw the title that the difficulties would be all over the place. That's why i did not say anything until you confirmed it. This is not something that can be easily automated and the completion rate alone can't even get close to the real picture.

In reality, setting difficulties is a very simple problem only if we ask the right questions. And this leaves no room for discussion because everyone is on the same page and it's very clear what needs to be done. That's why i was never interested in discussing this. I don't believe it's something that needs to be discussed. My intention was to present my approach and stop there. If there was an agreement, i would go over the exercises and figure out the new difficulties.

Discussing some things is ok. Discussing everything is not. People tend to disagree which means nothing ever gets done. I really wish there was an easier way for me to solve problems without going through all this, but at the same time, i don't want to force people into doing things they aren't really interested in.

Thumbs up to confirm that the discussion ends here.

PR #569, opened 2023-07-13
included conj in concepts/lists/about.md

Tesse0ract:


tasxatzial:
Wait a sec... The commit is:

`conj` returns a list with items appended in it.

This does not look correct. `conj` returns a list with the new item added to beginning. Works the same as `cons` except that `cons` returns a `Cons` object aka lazy seq, not list.

PR #574, opened 2023-07-17
Robot simulator: Sync with problem-specifications

### LINKED ISSUE #456: Tests in Robot Simulator are overly prescriptive
The robot simulator exercise contains two tests that cause problems for students.

```clojure
(deftest can-turn-right
  (is (= :north (robot-simulator/turn-right :west))))

(deftest can-turn-left
  (is (= :west  (robot-simulator/turn-left :north))))
```

A completely valid approach would be for `turn-left` and `turn-right` to accept a robot as input and produce a robot as output. However, the tests preclude this approach to solving the exercise. I would advocate for removing these tests.

tasxatzial:
Closes #456 

Complete re-write of tests. Took way longer than expected because I wasn't able to decide what's the most appropriate way to write them. So, i just picked the one that looked best to me.

### DIFF
--- a/exercises/practice/robot-simulator/.docs/instructions.md
+++ b/exercises/practice/robot-simulator/.docs/instructions.md
@@ -10,13 +10,10 @@ The robots have three possible movements:
 - turn left
 - advance
 
-Robots are placed on a hypothetical infinite grid, facing a particular
-direction (north, east, south, or west) at a set of {x,y} coordinates,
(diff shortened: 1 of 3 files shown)


tasxatzial:
Don't approve yet. Just open. There's an error i have to fix.

tasxatzial:
Ok, fixed, it was a just a deftest declared twice.

PR #575, opened 2023-07-23
Phone number: Sync with problem-specifications

### LINKED ISSUE #230: phone-number: README doesn't mention the functions area-code and pretty print
The README references the `number` function, but the others are only mentioned in the test suite.
I'll try to put together a PR for this in the next couple of days.

tasxatzial:
Closes #230 

Synced and non-required functions were removed from phone-number.clj

Tests were insufficient, so broken solutions are expected. The example solution still works fine though.

### DIFF
--- a/exercises/practice/phone-number/.docs/instructions.md
+++ b/exercises/practice/phone-number/.docs/instructions.md
@@ -2,9 +2,11 @@
 
 Clean up user-entered phone numbers so that they can be sent SMS messages.
 
-The **North American Numbering Plan (NANP)** is a telephone numbering system used by many countries in North America like the United States, Canada or Bermuda. All NANP-countries share the same international country code: `1`.
+The **North American Numbering Plan (NANP)** is a telephone numbering system used by many countries in North America like the United States, Canada or Bermuda.
+All NANP-countries share the same international country code: `1`.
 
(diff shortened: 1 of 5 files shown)


tasxatzial:
FWIW, I just realized that the current specs say that it's an error when the number is invalid. However, all tests use "0000000000" as the return value instead of throwing an error. This is something i kept from the old tests since i didn't want to break 100% of the solutions. Not that it would make any big difference. Currently 90% of the solutions are broken.

tasxatzial:
@bobbicodes When you have some time, i would really appreciate your input on this. Is it preferable to stick to the specs and throw exceptions? Or is it ok to return "0000000000"? Other languages like JS and python throw exceptions.

Also, i've noticed that syncing some exercises will break all the solutions. The tests in some of them are very outdated. Any idea how these should be dealt with?

bobbicodes:
I don't really have strong opinions about these things. I'm not opposed to breaking solutions but once you weigh all the factors like asking what will be gained by doing so and whether that will justify the time, I've concluded so far that the exercises are fine how they are unless there's something embarrassingly wrong or something. If you feel it is important I'm fine with merging stuff, but I don't feel I have much input here.

tasxatzial:
I agree, this exercise can stay as it is. Throwing exceptions isn't going to add any value.

PR #577, opened 2023-09-22
Armstrong numbers: Sync with problem-specifications

tasxatzial:
The updated tests from problem specifications require solutions that use big integers. While this can be easily done in Clojure, those tests were left out to avoid breaking existing solutions.


tasxatzial:
This one failed the tests for some reason. Needs further investigation.

tasxatzial:
The tests fail because the new tests are using larger numbers and the example solution should be updated to use `*'` instead of `*`. My main concern is that this will break many of the community solutions because they were not written with big integers in mind.

The other option is to update the tests and mark the failing tests as not implemented.

tasxatzial:
PR is ready. Tests that require big ints were marked as not implemented.

PR #589, opened 2023-10-18
Concepts/functions generating functions

JaumeAmoresDS:
Added functions-generating-functions concept


bobbicodes:
This looks really good so if you'd like, we could get the concept merged first. It looks like it's just missing the `.meta/config.json` file, and needs the concept added to the root `config.json` concepts array (with a fresh UUID).

The only feedback I have is that the examples might be easier to read without assigning them to vars, at least to my eye it would be cleaner that way... or maybe just add a newline or two to keep them from running too far to the right (the final rendered page stuffs them into a narrow code box, so I usually try to keep it from needing to be horizontally scrolled).

Thanks!

JaumeAmoresDS:
Many thanks @bobbicodes. I did as you suggested and added newlines for better readability. This option seemed to make it a bit more explicit the fact that we are generating new functions that later can be called with other arguments. Please let me know if there's still the issue of horizontally scrolling... (I can't see that in my markdown editor sorry!). 

I also added the .meta/config.json file and updated links.json. Thanks for your feedback!

concepts/functions-generating-functions/introduction.html
bobbicodes, line 1: Did you mean to include this file?

bobbicodes:
All that's left is to add this to `config.json` (after line 1090: https://github.com/JaumeAmoresDS/clojure/blob/45f137152262a48eed207c8798581f7a9a6a4b83/config.json#L1090):

```json
{
      "uuid": "9d9d7cf1-8504-4f7b-b2a0-1b3b638dc0d9",
      "slug": "functions-generating-functions",
      "name": "Functions generating functions"
 }
```

concepts/functions-generating-functions/about.md
bobbicodes, line 1: The about page should contain the same content as `introduction.md` (or more, but not less)

JaumeAmoresDS:
Great, many thanks for your help. I included the three changes you indicated. I just read about the configlet tool, and will make sure to include the UIUUD on a next occasion. Apologies for missing that!

bobbicodes:
This is great, thanks a lot!

JaumeAmoresDS:
Link to the discussion: http://forum.exercism.org/t/new-concept-deriving-new-functions-from-existing-functions-partial-comp-memoize/7746/6

PR #591, opened 2023-11-04
In exercises/concept/functions-generating-functions/, added:

JaumeAmoresDS:
.docs/instructions.md
   .docs/introduction.md
   .meta/exemplar.clj

### DIFF
--- a/exercises/concept/functions-generating-functions/.docs/instructions.md
+++ b/exercises/concept/functions-generating-functions/.docs/instructions.md
@@ -1,0 +1,283 @@
+# Instructions
+
+Mary and John like to do frequent trips by bike. Based on their previous trips, they have estimated that their average speed is `20 km/h`.
+
+```clojure
+(def john-mary-average-speed 20)
+```
+
+Based on their average speed, they want to estimate how long it will take to perform their next planned trips.
+
+# 1. Estimate total duration of a given trip
+
+Each trip is defined as a list of maps, for example:
+
+```clojure
+(def trip 
+  [{:origin "Paris"
+    :destination "Lyon"
+    :distance 800}
+  
+   {:origin "Lyon"
+    :destination "Milan"
+    :distance 500}
+  
+   {:origin "Milan"
+    :destination "Rome"
+    :distance 300}])
+
+```
+
+Using `map`, `reduce`, and `partial`, implement a function that takes as input the average speed and a given trip, and returns the estimated time that takes to do such trip:
+
+```clojure
+(defn estimated-trip-duration 
+  "Returns the time it takes to do the trip
+  given the average-speed."
+  [average-speed trip]
+  ;(reduce + (map my-partial-fn (:distance trip))))
+```
+
+in our case, if we run:
+
+```clojure
+(estimated-trip-duration john-mary-average-speed trip) ; => 80.0
+```
+
+we get 80.0
+
+In mathematical notation we have:
+ 
+```
+my-full-fn (speed, distance) => time it takes to travel given distance at given speed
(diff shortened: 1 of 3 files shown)


bobbicodes:
I like what you've done. I'm just a little bit concerned because we haven't taught map/reduce yet (or hashmaps for that matter!). If it weren't for this issue, I'd say it would fit perfectly, but it would leave the uninitiated student needing to learn these concepts elsewhere.

JaumeAmoresDS:
You're right @bobbicodes. Maybe we could just avoid suggesting to use `map` or `reduce`, so that the student can use any other mechanism of their choice, like loops, recursion, etc? Although, in all honesty, those haven't been introduced either... In my case, I applied `map` and `reduce` to solve other exercises in the `clojure` track. In one of the exercises, I applied recursion but my mentor suggested to use `reduce`, for instance.

JaumeAmoresDS:
@bobbicodes, on the other hand, I didn't fully understand the difficulty you mentioned about adding a concept for map and another for reduce. From what I know (very little, but something), adding basic information about map, for example, should be pretty straightforward: one or two sentences. Also, since it is not a complicated concept (at least not in how it's used most of the time), with just one or two exercises might be enough? 

Reduce, on the other hand, seems to have more use cases (but maybe just as juxt, since you even dedicated a blog to it?). However, I would dedicate multiple concepts that have a hierarchy of difficulty, something like this:

```mermaid
graph TD;
A{basic reduce} --> B{advanced reduce 1};
A{basic reduce} --> C{advanced reduce 2};
```

Unless you have something more elaborate in mind, I could volunteer on creating a **basic** concept for map, another for reduce, and then one or two concept exercises for each. I think I have bandwidth for doing that, but would need to wait for something more exhaustive... please let me know!

bobbicodes:
Map and reduce are high priority and would be a welcome addition, as they form the cornerstone of functional programming. I'm unlikely to contribute them myself because I've been busy working on other things lately, so if you were so inclined that would be wonderful :)

As far as this exercise goes, my gut reaction is that if it could be made simpler it would be more effective, in other words not requiring any additional strategies. Would that even be possible, or would it make the exercise fall apart? It's an almost obligatory suggestion because the learning exercises are meant to be just on the edge of being *excessively simple*, and when they are not, well that's when we start to receive feedback from frustrated students.

Of course if it were added *after* introducing other concepts it might be fine, because then it would only be introducing one new thing at a time.

JaumeAmoresDS:
Thanks @bobbicodes. If I understand correctly, I could first add map and reduce, and then we could have the current PR approved because solving the exercise would be much easier after map and reduce are in the track, is that correct? If you think I can submit a really simple explanation, one for map, and the other for reduce, and only one or two exercises for each one, then I can certainly do it!

bobbicodes:
Yeah, my most recent thoughts were to expand the syllabus significantly by taking a much simpler approach and making the exercises smaller, just like you're saying. More like along the lines of 4clojure or the Clojure Koans where it's right to the point.

JaumeAmoresDS:
That would be great, do you think then that we could be using the exercises from 4clojure directly? That would be a very fast way to cover many concepts.

bobbicodes:
Yeah, that's what I'm thinking! It's proven pedagogically sound, the problems were taken from various sources like Project Euler and Ninety-Nine Lisp Problems and organized into a well-ordered system, then a couple years ago the authors decided to let the domain expire. They said it had become outdated and they were no longer interested in maintaining it. It was revived as 4everClojure by the community but they have no ownership of the problem set, and I think it would be quite appropriate for them to be inherited by Exercism in some form.

JaumeAmoresDS:
Looking up 4clojure, I just read the list of websites dedicated to practicing clojure, in the practical.li site... so many nice possibilities!

https://practical.li/clojure/coding-challenges/

PR #595, opened 2023-11-23
Concept/map reduce

JaumeAmoresDS:
Added conceps for higher-order functions map and reduce


bobbicodes:
This is great, thanks! I'll probably only have minor suggestions

bobbicodes:
I'm thinking maybe the introduction pages could be made smaller. Since the purpose of having a separate Introduction and About page is for the former to contain *just enough* to get them started, while the About pages are more comprehensive.

(This is more of a statement on how thorough you made them than a criticism)

JaumeAmoresDS:
@bobbicodes would it be good like this? I removed the last two sections of each introduction.md, but kept them in about.md

bobbicodes:
Yes, that's perfect :)

bobbicodes:
While we are touching things related to these new concepts, it just occurred to me that the tree will be affected significantly due to their introduction. In other words, someone will need to go through the practice exercises in the root `config.json` and add map and/or reduce to the `prerequisites` and `practices` arrays.

This *could* be done in a later PR however.

JaumeAmoresDS:
I might be able to do it, but I think later in the future. The thing is that I am not familiar with the exercises, since I did very few so far. I'm also a newbie in clojure, which makes it a bit more challenging, although this might not be an issue if I am guided through the PR review process.

bobbicodes:
No worries, just thinking ahead :)

tasxatzial:
@JaumeAmoresDS Can you please use one sentence per line in the introduction.md and about.md of the reduce concept? It's common practice around here.

Only these two files please because i already have a review in progress about everything else.

tasxatzial (commented):
This is the first pass of the review. Includes everything except the `introduction.md` and `about.md` from the reduce concept.

@JaumeAmoresDS Please take a look and let me know if there are questions.

tasxatzial:
Given that you created this PR 1.5 year ago, please let me know if you are willing to go through couple rounds of review. If it's too much, I can merge right away and update it in a subsequent PR. No worries.

PR #610, opened 2024-01-22
Cars-assemble: fix tests failing due to floating point precision

tasxatzial:
A simple fix for the issue described below:

https://forum.exercism.org/t/cant-solve-cars-assemble-because-of-floating-point-precision/9242

The current tests check if two double numbers x, y are equal, and as a result some of them are failing when the numbers aren't identical. Two possible solutions: 
* Check if the two numbers are close enough using a check like `abs(x-y)<epsilon`.
* Truncate doubles to floats. This lowers the required precision required for the tests to pass.

This PR employs the second approach.

### DIFF
--- a/exercises/concept/cars-assemble/test/cars_assemble_test.clj
+++ b/exercises/concept/cars-assemble/test/cars_assemble_test.clj
@@ -2,29 +2,33 @@
   (:require cars-assemble
             [clojure.test :refer [deftest is testing]]))
 
+(defn- float-equal?
+  [x y]
+  (= (float x) (float y)))
+
 (deftest ^{:task 1} production-rate-speed-0-test
   (testing "Production rate for speed 0"
-    (is (= 0.0 (cars-assemble/production-rate 0)))))
+    (is (float-equal? 0.0 (cars-assemble/production-rate 0)))))
 
 (deftest ^{:task 1} production-rate-speed-1-test
   (testing "Production rate for speed 1"
-    (is (= 221.0 (cars-assemble/production-rate 1)))))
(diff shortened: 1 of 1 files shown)


tasxatzial:
I've just realized that my PR introduces a small typo . I'll correct it when it's reopened.

exercises/concept/cars-assemble/test/cars_assemble_test.clj
```diff
@@ -2,29 +2,33 @@
...
-    (is (= 221.0 (cars-assemble/production-rate 1)))))
+    (is (float-equal? 221.0 (cars-assemble/production-rate 1)))))
 
 (deftest ^{:task 1} production-rate-speed-4-test
   (testing "Production rate for speed 4"
-    (is (= 884.0 (cars-assemble/production-rate 4)))))
+    (is (float-equal? 884.0 (cars-assemble/production-rate 4)))))
 
-(deftest ^{:task 1} production-rate-speed-7-test
+(deftest ^{:task 1} production-rate-speeed-7-test
```
IsaacG, line 21: Is this change intentional? The validate exercise step should fail here.

tasxatzial: That's the typo i'm talking about.

tasxatzial:
@bobbicodes Do you have some time to review this one?

bobbicodes (approved):
Yeah I suppose this would fix a lot of problems with this exercise, which is not supposed to be a challenge, it *should* be simple at this point. I'll wait to merge it until the typo is fixed.

tasxatzial:
It's ready. Thanks!

PR #611, opened 2024-01-23
Removing code example with typo that was doubled in the coordinate-tr…

nicolas-lelouche:
…ansformation exercise


iHiD:
Thanks! :)

PR #662, opened 2024-09-06
say: Sync with problem specifications

### LINKED ISSUE #302: "say" exercise doesn't use "and" in hundreds
We explicitly have a test:
```clojure
(deftest one-hundred-twenty-three-test
  (is (= "one hundred twenty-three" (say/number 123))))
```

But in the description:

- 14 becomes "fourteen".
- 100 becomes "one hundred".
- 120 becomes "one hundred **and** twenty".
- 1002 becomes "one thousand **and** two".
- 1323 becomes "one thousand three hundred **and** twenty-three".

Seems are tests are much simpler (the examples conditionally add `and` between thousands or hundreds, depending on the value).

tasxatzial:
Also closes #302, as the 'extensions' section mentioned in #302 no longer exists in the synced version.

### DIFF
--- a/exercises/practice/say/.docs/instructions.md
+++ b/exercises/practice/say/.docs/instructions.md
@@ -6,11 +6,9 @@ Given a number from 0 to 999,999,999,999, spell out that number in English.
 
 Handle the basic case of 0 through 99.
 
-If the input to the program is `22`, then the output should be
-`'twenty-two'`.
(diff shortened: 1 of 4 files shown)


michalporeba (approved):
This change will potentially make some solutions to fail, so they should be regenrated, but at the same time we should merge it to conform to the problem specification.

PR #669, opened 2024-09-11
syncing roman-numerals to the problem specification

michalporeba:
Running sync on the roman numerals exercise to standardise the instructions and the test cases with the problem specification. Some of the test cases have changed so the solutions should be re-evaluated.


tasxatzial (commented):
The docs and metadata for all exercises were synced a few days ago, and there haven't been any changes to the problem descriptions since. It seems like your branch is out of sync, which could be causing the PR to show changes that don’t actually exist. Maybe do a `git rebase main` on your branch and then force-push?

tasxatzial (changes requested):
One test case is missing (1666), and the function `three-thousand` is defined three times.

tasxatzial:
Closing this as it is superseded by https://github.com/exercism/clojure/pull/734

PR #677, opened 2024-09-21
squeaky-clean: Update docs

tasxatzial:
This PR primarily fixes incorrect references to functions from the `clojure.string` namespace, such as `string/join`.

Additionally, I've edited and reworded certain parts to improve flow and maintain consistency.

Edit: In the third commit, the word 'need' was replaced to make it clearer that what follows are only suggestions.


iHiD (approved):
Approving to unblock

PR #679, opened 2024-09-26
annalyns-infiltration: Update docs

tasxatzial:
This PR primarily fixes an issue with an incorrect hint. It should explicitly state `true` instead of 'truthy'.

Edit: Changed 'that' to 'which', given the context (adding extra detail about something already clear)

Edit: Changed 'if' to 'whether', given the context (presenting two implied alternatives)


iHiD (approved):
Approving to unblock

PR #681, opened 2024-10-05
rna-transcription: Implement tests

tasxatzial:
Implementing tests to address the problem described here:

https://forum.exercism.org/t/confusing-message-about-a-test-failure-while-still-accepting-solution/13254


iHiD (approved):
Approving to unblock

PR #684, opened 2024-10-11
complex-numbers: Clarify that the exponent function isn't required

tasxatzial:
Adding an instructions.append to clarify that the exponent function isn't required.

Original thread:

https://forum.exercism.org/t/remove-mention-of-exponentiation-from-instructions/13225


iHiD (approved):
Approving to unblock

PR #685, opened 2024-10-12
Publish list-ops

tasxatzial:
Since the exercise deals with vectors, I've added it as extra prerequisite.

There's one more change needed (split into one deftest per test) so I'm marking it as draft.

### DIFF
--- a/config.json
+++ b/config.json
@@ -1367,8 +1367,7 @@
           "lists",
           "closures"
         ],
-        "difficulty": 5,
-        "status": "wip"
+        "difficulty": 5
       }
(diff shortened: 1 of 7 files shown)


iHiD (approved):
Approving to unblock

tasxatzial:
Not yet ready for merging, as the exercise can be approached in multiple ways. I’m currently exploring alternatives to enhance the learning experience.

tasxatzial:
Appendix has been expanded to include optional goals and extra suggestions.

Almost ready for publishing.

PR #686, opened 2024-10-15
Deprecate exercises

tasxatzial:
Since **'All your base'** has been implemented, the following exercises are marked as deprecated:
1. Binary
2. Trinary
3. Octal
4. Hexadecimal

I've also deprecated **'accumulate'** since **'list-ops'** will be published soon. Therefore, I'm marking the current PR as draft. It will be merged after https://github.com/exercism/clojure/pull/685


iHiD (approved):
Approving to unblock

PR #687, opened 2024-10-16
bird-watcher: Update exemplar solution

tasxatzial:
The hints for task 4 mention `every?`, `pos?`, and `not`. However, the exemplar solution is `(pos? (count (filter zero? birds)))`, which is not only different, but also suboptimal. This PR updates the solution to `(not (every? pos? birds))` to align with the hints.

I’ve also replaced the `last` function with `peek` since this exercise deals exclusively with vectors, and `peek` is the recommended way to retrieve the last element of a vector.

### DIFF
--- a/exercises/concept/bird-watcher/.docs/hints.md
+++ b/exercises/concept/bird-watcher/.docs/hints.md
@@ -6,7 +6,7 @@
 
 ## 2. Check how many birds visited today
 
-- The [`last`][last] function can be used to return the last item in a sequential collection.
+- The [`peek`][peek] function can be used to return the last item in a vector.
 
 ## 3. Increment today's count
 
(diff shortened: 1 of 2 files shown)


exercises/concept/bird-watcher/.docs/hints.md
```diff
@@ -6,7 +6,7 @@
 
 ## 2. Check how many birds visited today
 
-- The [`last`][last] function can be used to return the last item in a sequential collection.
+- The [`peek`][peek] function can be used to return the last item in a vector.
```
iHiD, line 9: People won't necessarily know what vectors are, so I think revert this, unless we've actively defined "vector" somewhere previosuly in the syllabus.

tasxatzial: > People won't necessarily know what vectors are, so I think revert this, unless we've actively defined "vector" somewhere previosuly in the syllabus.

This IS the learning exercise for vectors :)

https://exercism.org/tracks/clojure/concepts/vectors

iHiD (approved):

exercises/concept/bird-watcher/.docs/hints.md
iHiD: 👏

Issue #671, opened 2024-09-13
Descriptions and tests may be misleading regarding the type of both input and expected values


tasxatzial:
I'm opening a new issue to address and discuss two related, previously unresolved issues.

### Problem 1

Exercise descriptions will usually mention words that mislead new learners as to what type of input is expected. Here's an example from https://github.com/exercism/problem-specifications/blob/main/exercises/sublist/description.md

`If A = [1, 2, 3] and B = [], then A is a superlist of B`

Here, lists are mentioned, but vector notation is used. This is not only confusing but also appears incorrect.

This issue can also manifest in the descriptions of tests if the description is copied from the problem specifications. For example:

```
(deftest list-equals-itself-test
  (testing "list equals itself"
    (is (= :equal (sublist/classify [1 2 3] [1 2 3])))))
```

Also, the function name suggests that it accepts lists. Since we can pick whatever function name we want, the naming of the functions won't be discussed here but in another issue where we'll address a template for the tests.

Previous PR/issue: https://github.com/exercism/clojure/pull/507 https://github.com/exercism/clojure/issues/628

### Problem 2

Tests will typically use vectors to check for equality. Example:

```
(deftest list-equals-itself-test
    (is (= :equal (sublist/classify [1 2 3] [1 2 3]))))
```

Vectors are used as input arguments here, which may mislead new learners into thinking the input must be a vector, leading them to write solutions that work with vectors only.

Another example:

```
(deftest six
  (is (= [2, 3] (prime-factors/of 6))))
```

New learners may believe that the output should be a vector, when in fact, it could also be a list.

Previous PR/issue: https://github.com/exercism/clojure/issues/227

tasxatzial:
Current list of exercises that are affected

[all your base](https://github.com/exercism/clojure/tree/main/exercises/practice/all-your-base)
[allergies](https://github.com/exercism/clojure/tree/main/exercises/practice/allergies)
[anagram](https://github.com/exercism/clojure/tree/main/exercises/practice/anagram)
[binary search](https://github.com/exercism/clojure/tree/main/exercises/practice/binary-search)
[binary search tree](https://github.com/exercism/clojure/tree/main/exercises/practice/binary-search-tree)
[flatten array](https://github.com/exercism/clojure/tree/main/exercises/practice/flatten-array)
[grade school](https://github.com/exercism/clojure/tree/main/exercises/practice/grade-school)
[poker](https://github.com/exercism/clojure/tree/main/exercises/practice/poker)
[...]

PR #688, opened 2024-10-22
sublist: Sync with problem specifications

tasxatzial:
Also

* Added instructions append to clarify the input type
* Updated starter file sublist.clj (changed lists references to colls)


iHiD (approved):
Approving to unblock

tasxatzial:
Not yet ready for merging, as it falls under the category of exercises that use terminology conflicting with Clojure terminology. See #671

tasxatzial:
Ready for merging. I believe the append file is more than enough to address https://github.com/exercism/clojure/pull/507

We'll see how it goes and we'll revise in the future, if needed.

PR #693, opened 2024-10-28
allergies: Split into one deftest per test

tasxatzial:
Also

* Updated the allergies.clj starter file with proper function templates
* Added instructions append to clarify the output type


ErikSchierboom (approved):
Please merge with `[no important files changed]`

PR #694, opened 2024-10-31
Add Eliud's eggs exercise

kahgoh:
This is one of the [48in24 exercises](https://forum.exercism.org/t/48in24-exercise-11-05-eliuds-eggs/12900/3)

### DIFF
--- a/config.json
+++ b/config.json
@@ -350,6 +350,16 @@
         ]
       },
       {
+        "slug": "eliuds-eggs",
+        "name": "Eliud's Eggs",
+        "uuid": "bfe1789f-8388-4285-bed9-26f694e12c70",
+        "practices": [],
+        "prerequisites": [
+          "basics"
+        ],
+        "difficulty": 2
+      },
+      {
         "slug": "etl",
         "name": "ETL",
         "uuid": "befca7ee-fc29-4ef0-afd7-2d3f68aa47fb",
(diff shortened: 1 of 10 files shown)


tasxatzial (commented):
Would you mind moving its config entry after complex-numbers? There are two PRs open that modify the main config, and merging it as-is will create conflicts in both.

Also, adding 'basics' as a prerequisite would be nice to have.

tasxatzial (approved):

tasxatzial:
@kahgoh Since the instructions mention a `bit-count` function that doesn’t actually exist, I’ve added a small appendix to ensure there’s no confusion.

tasxatzial:
I see, so a solution that looks like so would be considered ok?

```
(defn egg-count [number]
  (->> (Integer/toBinaryString number)
       count
       range
       (map #(bit-test number %))
       (filter true?)
       count))
```

This converts to a binary string, constructs a list from 1 to the number of its digits, maps bit-test over that list to check if the bit at each index is true or false, and finally counts the number of true values.

tasxatzial:
Okay, I'll allow the `bit-*` functions for now and re-evaluate once community solutions start coming in.

tasxatzial:
I’m also thinking that 'don't use bit-count functionality' could easily be interpreted as 'don't use bit-* and count functions,' since bit-count isn't a Clojure function. I'll be monitoring the solutions closely, but I'll most likely add an appendix to prevent confusion.

PR #699, opened 2024-11-04
gigasecond: Sync tests

tasxatzial:
Also

* Updated the gigasecond.clj starter file with a proper function template
* Specs mention that some tests check only for date (not time), which appears to be the case in the existing implementation. Therefore, the tests that also check time were marked as excluded.
* Removed the commented out test since it's not part of the specs.


ErikSchierboom (approved):

PR #700, opened 2024-11-04
matching-brackets: Explicitly check with true? in tests

tasxatzial:
Made the tests consistent by explicitly checking with `true?` as well, as `true` is the other value expected to be returned by `valid?`

### DIFF
--- a/exercises/practice/matching-brackets/test/matching_brackets_test.clj
+++ b/exercises/practice/matching-brackets/test/matching_brackets_test.clj
@@ -2,82 +2,82 @@
   (:require [clojure.test :refer [deftest testing is]]
             matching-brackets))
 
-(deftest paired-square-brackets
-  (testing "paired square brackets"
-    (is (matching-brackets/valid? "[]"))))
+(deftest test-81ec11da-38dd-442a-bcf9-3de7754609a5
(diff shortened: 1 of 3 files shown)


ErikSchierboom (dismissed):

exercises/practice/matching-brackets/test/matching_brackets_test.clj
```diff
@@ -2,82 +2,82 @@
   (:require [clojure.test :refer [deftest testing is]]
             matching-brackets))
 
-(deftest paired-square-brackets
-  (testing "paired square brackets"
-    (is (matching-brackets/valid? "[]"))))
+(deftest test-81ec11da-38dd-442a-bcf9-3de7754609a5
```
ErikSchierboom, line 5: I'm not a fan of the use of the GUID here. Why the change

tasxatzial: Since no one seems to be sharing their opinions on the Clojure thread in the forum, I’ll create a new thread about function naming to gather input from other maintainers. I'd like to have a clearer understanding before reconsidering.

ErikSchierboom (approved):

PR #701, opened 2024-11-04
pangram: Sync tests

tasxatzial:
Also

* Updated the pangram.clj starter file with a proper function template
* Explicitly test with `true?` as well, as `true` is the other value expected to be returned by `pangram?`


ErikSchierboom (approved):

PR #705, opened 2024-11-09
isogram: Sync tests

tasxatzial:
Also

* Updated the isogram.clj starter file with a proper function template
* Split into one deftest per test
* Explicitly check with `true?` and `false?` as `true` and `false` are the exact values expected to be returned by `isogram?`
* Written a new example solution that passes the tests. It is expected that some of the community solutions will be invalided.


tasxatzial:
I'm not updating this; instead, I will close it if the isogram tests, which are currently auto-generated in the test generator PR, are updated in a new PR.

ErikSchierboom (approved):

PR #706, opened 2024-11-09
isbn-verifier: Sync tests

tasxatzial:
Also

* Updated the isbn_verifier.clj starter file with a proper function template
* Written a new example solution that passes the tests. It is expected that some of the community solutions will be invalided.
* Switch to idiomatic checking `(= true x)` -> `(true? x)`


ErikSchierboom (approved):

PR #710, opened 2024-11-24
list-ops: Update instructions append

tasxatzial:
* Standardize the first paragraph in the append file
* Update the wording in both goals

### DIFF
--- a/exercises/practice/list-ops/.docs/instructions.append.md
+++ b/exercises/practice/list-ops/.docs/instructions.append.md
@@ -8,7 +8,7 @@ This appendix provides additional clarification or modifies the instructions as 
 ~~~~
 
 For this exercise in the Clojure track, **assume both the input and output are vectors**, as indicated by the tests.
-As a stretch goal, consider how you could implement the solution without using lists anywhere in your code.
+As a stretch goal, consider how to implement an approach that does not use lists for intermediate steps.
 
 It is important not to reuse existing Clojure built-in functions with similar functionality, as doing so would diminish the intended learning value of the exercise.
 Key functions from the `clojure.core` namespace to avoid include `into`, `concat`, `cat`, `lazy-cat`, `mapcat`, `flatten`, `filter`, `filterv`, `remove`, `count`, `map`, `mapv`, `reduce`, `transduce`, `reverse`, and `rseq`.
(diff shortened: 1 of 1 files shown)


exercises/practice/list-ops/.docs/instructions.append.md
```diff
@@ -1,19 +1,27 @@
...
+
+It is important not to reuse existing Clojure built-in functions with similar functionality, as doing so would diminish the intended learning value of the exercise.
+Key functions from the **clojure.core** namespace to avoid include `into`, `concat`, `cat`, `lazy-cat`, `mapcat`, `flatten`, `filter`, `filterv`, `remove`, `count`, `map`, `mapv`, `reduce`, `transduce`, `reverse`, and `rseq`.
+
+### Optional goals
+
+Try to pass the tests by devising a solution that assumes:
+
+- both the input and output are lists instead of vectors
+- the test suite isn't modified
```
kotp, line 20: The test suite is not modified is a standard of Exercism.  Should we need to mention that here?

tasxatzial: I'm afraid I don't have any other option. The tests are already using vectors as input, and in the first bullet, I ask to assume that they are using lists. Without the second bullet, some folks might be confused, while others might believe that the only way the first assumption can be true is by downloading the exercise and manually changing the tests to use lists, which is definitely not intended. Both bullets are necessary to clarify what I’m explaining in line 22:

These assumptions directly influence the types of the functions that can be used.

This is an optional and slightly more challenging goal. I'm explicitly mentioning more than I usually would to make it a bit clearer how they should approach it without going off on tangents. Still, it assumes a certain level of maturity from the reader, and it's meant only for those who want extra practice.

In any case, the reason I marked this as a draft is to let it sit here while I think about whether I can find an alternative way to phrase the optional goal, or even remove it altogether.

kotp: We could state: `as usual, the tests are not meant to be modified`, since it is a confirmation that this is not different than other exercises, in this regard.

Otherwise, if I saw this fairly soon after coming to Exrecism, it would likely encourage me to change other exercise test files, which is also not intended.

tasxatzial: Sounds reasonable. I'll see how I can rephrase the assumptions. Thanks!

ErikSchierboom (approved):

PR #712, opened 2024-12-01
armstrong-numbers: Include excluded tests

tasxatzial:
Added the final two tests involving big integers. Without these tests, some solutions improperly relied on Java interop for exponentiation, which is incorrect since Java's exponentiation functions return floating-point numbers instead of integers, as required here. Additionally, without these tests, incorrect solutions using Java interop could pass all tests. This PR addresses both issues.

Also

* Updated the armstrong_numbers.clj starter file with a proper function template
* Explicitly check with `true?` and `false?` as `true` and `false` are the exact values expected to be returned by `armstrong?`
* Updated example solution to pass the new tests

### DIFF
--- a/exercises/practice/armstrong-numbers/.meta/config.json
+++ b/exercises/practice/armstrong-numbers/.meta/config.json
@@ -7,7 +7,8 @@
     "imtayadeway",
     "mawis",
     "mtkp",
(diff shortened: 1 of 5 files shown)


ErikSchierboom (approved):
Nice!

PR #716, opened 2025-01-05
Add script to scaffold new practice exercises

ErikSchierboom:


bin/add-practice-exercise
```diff
@@ -0,0 +1,90 @@
...
+sed -i "s/acronym/${slug}/g" "${exercise_dir}/project.edn"
+
+cat << NEXT_STEPS
+
+Your next steps are:
+- Create the test suite in $(jq -r '.test' <<< "${files}")
+  - The tests should be based on the canonical data at 'https://github.com/exercism/problem-specifications/blob/main/exercises/${slug}/canonical-data.json'
+  - Any test cases you don't implement, mark them in 'exercises/practice/${slug}/.meta/tests.toml' with "include = false"
+- Create the example solution in $(jq -r '.example' <<< "${files}")
+- Verify the example solution passes the tests by running 'bin/verify-exercises ${slug}'
```
tasxatzial, line 86: This refers to a bin/verify-exercises script that appears to be missing?

ErikSchierboom: Ah. I'll update! Well spotted

ErikSchierboom: See https://github.com/exercism/clojure/pull/717

ErikSchierboom:
I've also updated the PR to write the namespaces to the solution, example and test files.

tasxatzial:
So is the scaffolding supposed to run only on linux? I tried it on windows with 'git bash' and it works fine. Only verifying the exercise seems to fail.

ErikSchierboom:
> So is the scaffolding supposed to run only on linux? I tried it on windows with 'git bash' and it works fine. Only verifying the exercise seems to fail.

For Windows, I would recommend running it on WSL. Most tracks are maintained by users that are on Linux, or else have access to WSL.

tasxatzial:
Thanks, i'll give WSL a go. This will take some time given it's the first time i have to look at anything but exercises. So I'd like to do this properly, but if you are in a hurry, please go ahead and merge.

tasxatzial (approved):

PR #719, opened 2025-01-05
Refactor test.clj

ErikSchierboom:


tasxatzial (approved):
This appears to be a simple refactoring. No change in functionality. Correct?

PR #728, opened 2025-01-09
Add test generator

ErikSchierboom:
@tasxatzial This is a very basic test generator prototype. The way it works is that you create a `generator.template` file in the exercise's `.meta` directory and then run `bin/generate-tests <slug>`. This then:

1. Fetches the `problem-specifications` repo locally (or pulls the latest when it is already there)
2. Parses the exercise's `canonical-data.json` file to find the test cases
3. Parses the exercise's `tests.toml` file to exclude test cases that are set to `include = false` in the `tests.toml` file (this happens automatically for reimplemented test cases)
4. Renders the `generator.template` file using the fetched test cases

The generator itself is located in the `generators` directory and uses a `deps.edn` file to manage its dependency.
The generator doesn't yet support any pre-processing of canonical data, but I imagined that to be done via an optional `generator.clj` file in the `.meta` directory.


exercises/practice/collatz-conjecture/.meta/tests.clj
ErikSchierboom, line 1: This is auto-generated

ErikSchierboom:
I'll review the actual templates from the PR before we'll merge, as they should be added in separate PRs

exercises/practice/roman-numerals/test/roman_numerals_test.clj
```diff
@@ -1,57 +1,112 @@
 (ns roman-numerals-test
-  (:require [clojure.test :refer [deftest is]]
-            roman-numerals))
+  (:require [clojure.test :refer [deftest testing is]]
+             roman-numerals))
 
-(deftest one
-  (is (= "I" (roman-numerals/numerals 1))))
+(deftest roman-numerals_test_1
```
ErikSchierboom, line 5: If the tested function has dashes in it (like `roman-numerals`), do we want to keep that in the function name or would you rather replace it with underscores?

ErikSchierboom:
I'll check to see if there is a nice handlebars library, as I think we'll be needing conditionals

exercises/practice/roman-numerals/test/roman_numerals_test.clj
tasxatzial: I don't have any strong preference. We can keep them to indicate more clearly the tested function, and use underscores for the _test_1 part. Or, we can use dashes everywhere so that we follow the language conventions. Or, we can use all underscores. Do you have any preference or suggestion?

ErikSchierboom: I don't. I'm happy to go with what you prefer

ErikSchierboom:
I'm switching to hbs for handlebars compatible templates which are nicer to work with and extensible. I'll do some work on more involved exercises

exercises/practice/roman-numerals/test/roman_numerals_test.clj
tasxatzial: Let's go with your suggestion then. Dashes for the function name, underscores for the _test_1 part.

tasxatzial: I'll also fix the names in my open PRs. Should be trivial to update them all. Will notify when they are done.

tasxatzial:
> I'll review the actual templates from the PR before we'll merge, as they should be added in separate PRs

If you also post the isogram tests in a new PR, i'll close https://github.com/exercism/clojure/pull/705

generators/src/templates.clj
```diff
@@ -0,0 +1,42 @@
...
+
+(defn- transform [slug test-cases]
+  (let [transform-file (paths/generator-clojure-file slug)]
+    (if (.exists transform-file)
+      (let [generator-ns (symbol (str slug "-generator"))]
+        (load-file (str transform-file))
+        (if-let [transform-fn (ns-resolve generator-ns (symbol "transform"))]
+          (transform-fn test-cases)
+          test-cases))
+      test-cases)))
```
ErikSchierboom, line 30: This now allows for a `generator.clj` file next to the `generator.template` file that can be used to customize the data before it is passed to the template.

ErikSchierboom:
I've happy with what I have now, and it should be enough for a lot of exerciss.

ErikSchierboom:
Open question before we'll merge PRs regarding the template: do we want `generator.template` or `generator.tpl` files?

exercises/practice/roman-numerals/test/roman_numerals_test.clj
tasxatzial: @ErikSchierboom All my open PRs are finished, but we both have PRs for pangram and difference-of-squares. Take a look at the pangram PR to see how the tests should be written in that case. We should be explicitly checking for true or false return values. The reason for it is because without explicit checks, someone can pass the tests without actually returning true or false. And given the function ends in ?, we are expecting only true and false to be returned.

ErikSchierboom: I'll go through your PRs later and approve them. Once they're merged I'll update my PRs to just add the template.

ErikSchierboom: And I agree with `true?` and `false?`

ErikSchierboom:
I think I'll go with `.tpl`

generators/src/templates.clj
```diff
@@ -0,0 +1,42 @@
...
+  (:require [hbs.core :as hbs]
+            [clojure.string :as str]
+            [log]
+            [paths]))
+
+(def exercises-with-template
+  (->> paths/exercises-dir
+       (file-seq)
+       (filter #(.isFile %))
+       (filter #(= "generator.template" (.getName %)))
```
tasxatzial, line 11:
```suggestion
       (filter #(= "generator.tpl" (.getName %)))
```

ErikSchierboom:
@tasxatzial should be good now!

ErikSchierboom:
@tasxatzial should be good now!

generators/src/templates.clj
ErikSchierboom: Forgot to push this change 🤦

ErikSchierboom:
@tasxatzial I'll wait for this PR to be merged and then I'll double check all the other PRs

bin/add-practice-exercise
```diff
@@ -96,11 +99,28 @@ for file_type in solution example; do
...
-cat << NEXT_STEPS
-Your next steps are:
+if [[ -f "${prob_specs_dir}/exercises/${slug}/canonical-data.json" ]]; then
+    cp "${exercise_dir}/${test_file}" "${generator_tpl_file}"
+
+    TEST_STEPS=$(cat << EOF
+- Create the test generator in '${generator_tpl_file}'
+    - The test generator uses the canonical data from 'https://github.com/exercism/problem-specifications/blob/main/exercises/${slug}/canonical-data.json'
+  - Any test cases you don't want to implement, mark them in 'exercises/practice/${slug}/.meta/tests.toml' with "include = false"
+  - Run 'bin/generate-exercise-tests ${slug}' to generate the tests
```
ErikSchierboom, line 109:
```suggestion
  - Run 'bin/generate-tests ${slug}' to generate the tests
```

tasxatzial:
The sum of multiples template has been added here accidentally.

tasxatzial:
> @tasxatzial I'll wait for this PR to be merged and then I'll double check all the other PRs

Ok, i'll do it as well.

ErikSchierboom:
> The sum of multiples template has been added here accidentally.

Fixed!

tasxatzial (approved):
Ok, let's proceed then. I believe this is good to go. Feel free to merge.

PR #730, opened 2025-01-11
atbash-cipher: Sync tests with problem specifications

tasxatzial:
Last two decode tests have been marked as not implemented because the inputs are not valid encoded strings. They should be removed from the specs as well.

https://forum.exercism.org/t/atbash-cipher-are-the-last-two-tests-invalid/13482

Also, this update will invalidate all solutions, as the old tests did not even include the decoding part of the exercise.


ErikSchierboom (approved):

ErikSchierboom:

PR #736, opened 2025-01-11
pangram: update to latest 

ErikSchierboom:


exercises/practice/pangram/.meta/generator.tpl
```diff
@@ -0,0 +1,13 @@
+(ns pangram-test
+  (:require [clojure.test :refer [deftest testing is]]
+             pangram))
```
tasxatzial, line 3:
```suggestion
            pangram))
```

exercises/practice/pangram/.meta/generator.tpl
```diff
@@ -0,0 +1,13 @@
...
+
+{{#test_cases.isPangram~}}
+(deftest pangram?_test_{{idx}}
+  (testing "{{description}}"
+    {{#expected~}}
+    (is (pangram/pangram? "{{input.sentence}}"))))
```
tasxatzial, line 9:
```suggestion
    (is (true? (pangram/pangram? "{{input.sentence}}")))))
```

exercises/practice/pangram/.meta/generator.tpl
```diff
@@ -0,0 +1,13 @@
...
+    {{else~}}
+    (is (not (pangram/pangram? "{{input.sentence}}")))))
```
tasxatzial, line 11:
```suggestion
    (is (false? (pangram/pangram? "{{input.sentence}}")))))
```

tasxatzial (commented):
* .problem-specifications directory is included

The conflict can be resolved by keeping the old version of the file since it was updated in one of my recent PRs.

ErikSchierboom:
@tasxatzial Apparently there is some quoting being done when rendering. The underlying libraries has an escaping option: https://github.com/jknack/handlebars.java/blob/master/handlebars/src/main/java/com/github/jknack/handlebars/EscapingStrategy.java, but I don't yet know how to tweak that from the Clojure library. Converting this to a draft.

ErikSchierboom:
I've managed to fix it. I've also introduced a `string` helper that I will probably want to use in the other templates too, as it escapes characters in strings.

tasxatzial (approved):
Nice! The generated tests now match my version.

PR #743, opened 2025-01-13
Add `high-scores` exercise

ErikSchierboom:


tasxatzial:
This is one of the exercises where the terminology from canonical data conflicts with Clojure terminology. There's an open issue about it here (no discussion, though) because there have complaints that it's confusing. See https://github.com/exercism/clojure/issues/671

To sum up, scores_test_1 and personal-top-three_test_1 both use the word "list", but the actual test uses vectors. One option is to change their descriptions to refer to vectors. Another option is to do nothing. But in any case, I'm always adding an append file because the description also uses lists.
This one here can be copied as is:https://github.com/exercism/clojure/blob/main/exercises/practice/sublist/.docs/instructions.append.md

ErikSchierboom:
@tasxatzial I've converted the template and tests to use a list

tasxatzial (commented):
Given that we normally use vectors to store data, lists are my least preferred option. But that's ok, it's a simple exercise, I guess it's ok to use lists.

ErikSchierboom:
I thought that it was perhaps good to also have students work with lists occasionally. I you feel that vectors would be better I'm happy to revert and add the append file.

tasxatzial (approved):

Issue #670, opened 2024-09-13
How to handle input validation and other exceptions in exercises?


michalporeba:
## Context 

Some exercises require input validation. A [simple example](https://github.com/exercism/problem-specifications/blob/578f37a3bde48be53726747c9ce3158102bb2880/exercises/grains/canonical-data.json#L85C9-L95C10) (copied below) from the [grains exercise](https://github.com/exercism/problem-specifications/tree/578f37a3bde48be53726747c9ce3158102bb2880/exercises/grains).

```json
{
      "uuid": "61974483-eeb2-465e-be54-ca5dde366453",
      "description": "negative square is invalid",
      "property": "square",
      "input": { "square": -1 },
      "expected": { "error": "square must be between 1 and 64" }
}
```

At the moment, the Clojure track lacks a consistent approach to what to expect from learners and, therefore, how to implement the tests. 

A **consistent** approach is important to allow learners to focus on the core of the exercise, on the concept which is being presented without needing to learn various ways of error handling that are possible in Clojure. Error handling should be explored with specific examples in an error-handling concept. 

For the same reason, we should choose a **simple** or **the simplest** approach. 

But it should also be **idiomatic** so that we are not introducing practices which are not natural for the language. 

### Option 1 - return error value

Some of the current exercises return an "error value", for [example](https://github.com/exercism/clojure/blob/327a9de9cf22dd291803be566bcc65f50b3749d3/exercises/practice/phone-number/test/phone_number_test.clj#L37) in the [phone number](https://github.com/exercism/clojure/tree/main/exercises/practice/phone-number) exercise have:

```clojure
(deftest invalid-with-letters
  (testing "Invalid with letters"
    (is (= "0000000000" (phone/number "523-abc-7890")))))
```

This could be standardised to include the expected description of the error:

```clojure
(deftest invalid-with-letters
  (testing "Invalid with letters"
    (is (= "error: phone number is invalid with letters" (phone/number "523-abc-7890")))))
```

**Pros**: The implementation of the `phone/number` could be left to the learner. Any flow control would do. 
**Cons**: It is generally not a good practice as error handling might be difficult. 

### Option 2 - asserts

A relatively common pattern in Clojure is to use `assert`s. An example of testing could be the following:

```clojure
(deftest test-zero-guard
  (testing "error handling test"
    (is (thrown? AssertionError (divide 2 0)))))
```

or with the exact message

```clojure
(deftest test-zero-guard
  (testing "error handling test"
[...]

PR #745, opened 2025-01-13
wordy: update to latest

ErikSchierboom:
This demonstrates how easy it is to handle errors

### DIFF
--- a/exercises/practice/wordy/.meta/generator.tpl
+++ b/exercises/practice/wordy/.meta/generator.tpl
@@ -1,0 +1,12 @@
+(ns wordy-test
+  (:require [clojure.test :refer [deftest testing is]]
+            wordy))
+{{#test_cases.answer}}
+(deftest evaluate_test_{{idx}}
+  (testing {{description}}
+    {{~#if error}}
+    (is (thrown-with-msg? IllegalArgumentException #{{error}} (wordy/evaluate {{input.question}})))))
+    {{else}}
+    (is (= {{expected}} (wordy/evaluate {{input.question}})))))
+    {{/if~}}
+{{/test_cases.answer~}}
--- a/exercises/practice/wordy/src/wordy.clj
+++ b/exercises/practice/wordy/src/wordy.clj
@@ -1,5 +1,7 @@
 (ns wordy)
 
-(defn evaluate [] ;; <- arglist goes here
-      ;; your code goes here
-)
+(defn evaluate
+  "Evaluate a simple math problem"
+  [question]
+  ;; function body
+  )
--- a/exercises/practice/wordy/src/wordy.clj
+++ b/exercises/practice/wordy/src/wordy.clj
@@ -1,7 +1,7 @@
 (ns wordy)
 
 (defn evaluate
-  "Evaluate a simple math problem"
+  "Evaluates a simple math problem"
   [question]
   ;; function body
   )
--- a/exercises/practice/wordy/.meta/config.json
+++ b/exercises/practice/wordy/.meta/config.json
@@ -6,7 +6,8 @@
     "AndreaCrotti",
     "haus",
     "sjwarner-bp",
-    "yurrriq"
+    "yurrriq",
+    "erikschierboom"
   ],
   "files": {
     "solution": [
--- a/exercises/practice/wordy/.meta/example.clj
+++ b/exercises/practice/wordy/.meta/example.clj
@@ -11,13 +11,22 @@
 
 (defn- parse-op [op-str]
   (or (ops op-str)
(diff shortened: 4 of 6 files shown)


tasxatzial:
Nice. The error handling is an open and stale issue atm. See https://github.com/exercism/clojure/issues/670

The reason I haven't yet synced exercises that include errors is that there isn't a consensus on the topic. There are many ways to handle errors, but my personal preference is to throw some kind of exception, usually either a generic Exception or an IllegalArgumentException. However, the error messages from the canonical data should probably be incorporated into the tests. For instance, this exercise includes two distinct error messages: 'unknown operation' and 'syntax error.' It is expected that participants should write code that can distinguish between these messages.

The main question, then, is: What kind of exceptions should be returned for each message?

ErikSchierboom:
> The main question, then, is: What kind of exceptions should be returned for each message?

Good question. I don't have much Java experience so I'm more than happy to go with what you'd like.

tasxatzial:
Then we can keep the IllegalArgumentException for all errors and also include the error message in the tests.

ErikSchierboom:
I'll need to fix the example implementation. Do the tests look good to you?

exercises/practice/wordy/.meta/generator.tpl
```diff
@@ -0,0 +1,12 @@
+(ns wordy-test
+  (:require [clojure.test :refer [deftest testing is]]
+            wordy))
+{{#test_cases.answer}}
+(deftest evaluate_test_{{idx}}
+  (testing {{string description}}
+    {{~#if error}}
+    (is (thrown-with-msg? IllegalArgumentException #{{string error}}` (wordy/evaluate {{string input.question}})))))
```
tasxatzial, line 8: There's an extra backtick here (doesn't seem to affect the test results)

ErikSchierboom:
@tasxatzial I've fixed the example implementation, although I feel like the canonical data is slightly weird. In particular:

```json
{
      "uuid": "8a7e85a8-9e7b-4d46-868f-6d759f4648f8",
      "description": "Non math question",
      "property": "answer",
      "input": {
        "question": "Who is the President of the United States?"
      },
      "expected": {
        "error": "unknown operation"
      }
    },
```

why isn't that a syntax error?

This test case makes sense for an unknown operation:

```json
{
      "uuid": "3ad3e433-8af7-41ec-aa9b-97b42ab49357",
      "description": "unknown operation",
      "property": "answer",
      "input": {
        "question": "What is 52 cubed?"
      },
      "expected": {
        "error": "unknown operation"
      }
    },
```

FYI It is fine to deviate/tweak the canonical data if we feel that it benefits the exercise. We should at least consider that here I think

tasxatzial:
Agree, this looks that it should have been a syntax error. So this is a case for a manual tweak?

ErikSchierboom:
Indeed. I'll update

ErikSchierboom:
Updated

exercises/practice/wordy/src/wordy.clj
```diff
@@ -1,5 +1,7 @@
 (ns wordy)
 
-(defn evaluate [] ;; <- arglist goes here
-      ;; your code goes here
-)
+(defn evaluate
+  "Evaluate a simple math problem"
```
tasxatzial, line 4:
```suggestion
  "Evaluates a simple math problem"
```

tasxatzial (approved):
Tested and seems to be working fine.

There's a sort of weird situation going on with the error handling in tests, that I feel is worth noting. The tests will pass only if the error thrown is a string that contains the string found in the tests.

Also, calling Character/isDigit without wrapping it in anonymous function is a new feature in Clojure 1.12.0. @ErikSchierboom I'm thinking of bumping leiningen configs to 1.12.0 to allow a bit more flexibility if someone is using it locally.

ErikSchierboom:
> The tests will pass only if the error thrown is a string that contains the string found in the tests.

I've made the tests more specific, where it has to match the string exactly.

> I'm thinking of bumping leiningen configs to 1.12.0 to allow a bit more flexibility if someone is using it locally.

Sounds good!

tasxatzial:
> I've made the tests more specific, where it has to match the string exactly.

The error messages is a topic I raised in discord to get an idea what people would do if the message is modified for some reason. Here's the thread https://discord.com/channels/854117591135027261/1303832856534974464/1303832859815186474

What do you think the best approach would be regarding the check for the error messages?

tasxatzial:
> I value people using the right error message for their errors enough that we should probably check for it.

No problem. If the message changes the regex can be adjusted to avoid invalidating solutions.

PR #759, opened 2025-01-14
Improve rendering using multimethod

ErikSchierboom:


ErikSchierboom:
@tasxatzial I've dug a bit deeper into the handlebars.java library and found that one could add custom formatters. With these, it suddenly becomes a lot easier to "just render things," without having to call special helper functions in the template.

tasxatzial:
Nice. Is it possible to clean up this PR a bit? It seems to be changing too many files. Perhaps we could consider leaving only the changes to the templates.clj file here, and move all the other changes to a separate PR.

Also, the tests in sum-of-multiples look that they use the reverse lists? It's probably not going to cause any issues, but the order is something that an existing solution might have taken into account.

ErikSchierboom:
@tasxatzial Agreed! I've updated the PR.

tasxatzial (approved):
Seems to be working fine.

PR #762, opened 2025-01-15
Add `saddle-points` exercise

ErikSchierboom:


tasxatzial:
The tests here expect the user to return the points in any order. Is this something we want, or should we let the user figure out that they should be returning the results in a specific order? What do you think would be the best approach?

There's also the option of using sets, if we want the user to get some practice on sets.

ErikSchierboom:
I don't want to force the student to return a specific order, they're tasked with finding the saddle points, and in which order that is doesn't matter. 

I'm fine with using a set here, but we should then maybe also add an append file?

tasxatzial:
> I'm fine with using a set here, but we should then maybe also add an append file?

What would the append file say?

ErikSchierboom:
That we expect them to return a set? Or do you think that's not necessary.

tasxatzial:
I usually add an append file when there's some part of the description that conflicts with the tests (if the description talks about lists and the tests use vectors, for instance). In every other case, an append file can be added, but doesn't really achieve much given that users will be checking the tests in order to figure out what they should be returning.

Shall we go with sets?

ErikSchierboom:
@tasxatzial Agreed! Updated.

tasxatzial (approved):
Looks good

PR #765, opened 2025-01-15
Simplify templates

ErikSchierboom:
Once this is merged, I'll do a PR to change `transform` to `transform-test-case` where applicable


tasxatzial:
I'll review saddle points, dnd, and wordy so that these are merged first. Then this PR can be updated.

tasxatzial (approved):
I see no issues with this. I'll approve so that you can update the rest of the open PRs, if needed.

PR #767, opened 2025-01-15
Add `zebra-puzzle` exercise

ErikSchierboom:


ErikSchierboom:
@tasxatzial Do you have any idea why this fails? If I copy the example file over the solution file locally and then run `clj -X:test` it works. Maybe it babashka?

ErikSchierboom:
Apparently it is a babashka thing:

> ; java.lang.IllegalArgumentException: No matching method indexOf found taking 1 args for class clojure.lang.Cons user /Users/erik/Code/exercism/tracks/clojure/exercises/practice/zebra-puzzle/.meta/example.clj:20:19

exercises/practice/zebra-puzzle/.meta/example.clj
```diff
@@ -0,0 +1,56 @@
...
+(defn- permutations [s]
+  (lazy-seq
+   (if (next s)
+     (for [head s
+           tail (permutations (disj s head))]
+       (cons head tail))
+     [s])))
+
+(defn index-of [coll e]
+  (first (keep-indexed #(when (= e %2) %1) coll)))
```
ErikSchierboom, line 18: This is a workaround for babashka weirdness

tasxatzial: I'm not surprised. Babashka implements a subset of Clojure and certain Java interop is left out. Apparently calling .indexOf on lazy sequences isn't possible.

exercises/practice/zebra-puzzle/.meta/generator.clj
```diff
@@ -0,0 +1,6 @@
+(ns zebra-puzzle-generator
+  (:require [hbs.helper :refer [safe-str]]
+            [clojure.string :as str]))
+
+(defn transform-test-case [test-case]
+  (update test-case :expected #(safe-str (str/lower-case (str ":" %)))))
```
tasxatzial, line 6: Another option is to use the function `keyword` if we want to make it more clear
```suggestion
  (update test-case :expected #(safe-str (str/lower-case (keyword %)))))
```

tasxatzial (approved):
Props for solving this!

PR #770, opened 2025-01-16
largest-series-product: Sync tests

tasxatzial:
Taking a stab at this. One thing that concerns me a bit is the type of Exception thrown. Right now all of them are marked as IllegalArgument, but perhaps not all of them should. In particular, a span that has length 5 and a string "123" will throw Exception even both of them are valid as individual arguments. It's just the combination that isn't valid.

@ErikSchierboom Any ideas?

### DIFF
--- a/exercises/practice/largest-series-product/.meta/config.json
+++ b/exercises/practice/largest-series-product/.meta/config.json
@@ -11,7 +11,8 @@
     "mathias",
     "petertseng",
     "rubysolo",
-    "yurrriq"
+    "yurrriq",
+    "tasxatzial"
   ],
   "files": {
     "solution": [
--- a/exercises/practice/largest-series-product/.meta/example.clj
+++ b/exercises/practice/largest-series-product/.meta/example.clj
@@ -1,15 +1,21 @@
 (ns largest-series-product)
 
-(defn- char->digit [c]
-  {:pre [(Character/isDigit c)]}
-  (Character/digit c 10))
-
-(defn- digits [ds] (map char->digit ds))
-
-(defn- slices [n ds] (partition n 1 (digits ds)))
-
-(defn largest-product [size ds]
+(defn check-input
+  [n digits]
   (cond
-    (zero? size)        1
-    (> size (count ds)) (throw (Exception. "Span must not exceed length."))
-    :else               (apply max (map (partial apply *) (slices size ds)))))
+    (neg? n) (throw (IllegalArgumentException. "span must not be negative"))
+    (or (> n (count digits)) (and (pos? n) (empty? digits))) (throw (IllegalArgumentException. "span must be smaller than string length"))
+    (and (pos? n) (empty? digits)) (throw (IllegalArgumentException. "span must be smaller than string length"))
+    (not-every? #(<= 0 % 9) digits) (throw (IllegalArgumentException. "digits input must only contain digits"))
+    :else false))
+
+(defn largest-product
+  [n s]
+  (let [digits (map #(Character/digit ^char % 10) s)]
+    (or (check-input n digits)
+        (if (zero? n)
+          1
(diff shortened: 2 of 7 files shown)


ErikSchierboom:
> In particular, a span that has length 5 and a string "123" will throw Exception even both of them are valid as individual arguments. It's just the combination that isn't valid.

I've checked the built-in exceptions and I don't think there is a better exception.

tasxatzial:
Ok, it's ready for review then.

ErikSchierboom (approved):
Looks good! I hope the templating wasn't that much of a hassle.

tasxatzial:
I'm mostly copying and pasting while editing older templates. However, I'm not comfortable with generator.clj files, as I'm still confused about their inner workings. Everything moves faster than I'm able to keep up with.

ErikSchierboom:
> I'm mostly copying and pasting while editing older templates. 

To be fair, that's also what I do :)

> However, I'm not comfortable with generator.clj files, as I'm still confused about their inner workings. 

They're sort of like optional plug-in points, where if the file exists _and_ it defines a `transform-test-test-cases` and/or `transform-test-test-case` function, those functions get called with the canonical data fetched from the prob-specs repo. This step happens _before_ the templates are rendered. I've tried to document this in https://exercism.org/docs/tracks/clojure/generators, but let me know if that is insufficient.

> Everything moves faster than I'm able to keep up with.

Sorry about that! I quite used to moving very fast, having been part of the Exercism team for 5 years and having been a (mostly autonomous) maintainer for the C#/F#/Prolog tracks.

tasxatzial:
> I've tried to document this in https://exercism.org/docs/tracks/clojure/generators, but let me know if that is insufficient.

I think i've gotten the general idea, i'm just unsure how to use it. It's probably going to be simple once i get the details. The docs have two functions, with very similar names, one updates a test case, and the other adds/removes? I've also looked at the source code, but this did not help much either. At this point I'm not even sure if there's any kind of error hidden in either the docs or the code.

> Sorry about that! I quite used to moving very fast, having been part of the Exercism team for 5 years and having been a (mostly autonomous) maintainer for the C#/F#/Prolog tracks.

That's ok. I aware that i can't possibly match the expertise of most maintainers. I'm just slow :)

ErikSchierboom:
> I've also looked at the source code, but this did not help much either. At this point I'm not even sure if there's any kind of error hidden in either the docs or the code.

Why makes you say there is an error in the docs or code?

The starting code is:

```clojure
(defn- transform [slug test-cases]
  (let [transform-file (paths/generator-clojure-file slug)]
    (if (.exists transform-file)
      (let [generator-ns (symbol (str slug "-generator"))]
        (load-file (str transform-file))
        (->> test-cases
             (transform-all-test-cases generator-ns)
             (transform-individual-test-cases generator-ns)))
      test-cases)))
```

This checks for a `exercises/practice/<slug>/.meta/generator.clj` file, and if that exists, it loads that file.
It also defines the expected namespace (`<slug>-generator`) that we expect that file to declare.
Then `transform-all-test-cases` and `transform-individual-test-cases` care called, both of which contain a bit of code to check if the just mentioned namespace contains such a named function, and if so, calls that function.

> The docs have two functions, with very similar names, one updates a test case, and the other adds/removes?

This is a very valid point, and maybe we should be renaming the functions to be more descriptive?
Maybe `add-remove-test-cases` and `update-test-case` would make more sense?

tasxatzial:
> Why makes you say there is an error in the docs or code?

Probably the similar function names, that we haven't used the add/remove function, and that both sections in the docs start with "To update individual test cases, define the following function:"

> Maybe add-remove-test-cases and update-test-case would make more sense?

Sure :)

ErikSchierboom (approved):

PR #773, opened 2025-01-17
Add `resistor-color-duo` exercise

ErikSchierboom:


exercises/practice/resistor-color-duo/src/resistor_color_duo.clj
```diff
@@ -0,0 +1,7 @@
+(ns resistor-color-duo)
+
+(defn value
```
tasxatzial, line 3: Should we be more specific here? `value` seems to be a bit too generic for a function name.
```suggestion
(defn resistor-value
```

tasxatzial (commented):
Let

exercises/practice/resistor-color-duo/src/resistor_color_duo.clj
ErikSchierboom: Fair point!

tasxatzial:
When you have the time, please also update the template and example files to use the new function name `resistor-value`.

ErikSchierboom:
Sure! Updated.

tasxatzial:
I can only see the changes if i click on the force-push message. But when i click on Files Changed tab, it still shows the old version. Does this need some kind of rebase?

ErikSchierboom:
Try again?

tasxatzial (approved):
Looks ok now 👍

PR #775, opened 2025-01-17
Improve naming of extension functions

ErikSchierboom:


docs/GENERATORS.md
```diff
@@ -65,15 +65,15 @@ This example removes all but the last element of the `:path` value (shortening t
 To update individual test cases, define the following function:
```
tasxatzial, line 65: Looks good. Only this line needs to be updated.

tasxatzial: Could you also update line 10 in the docs. There are couple extra [ ] there.

ErikSchierboom:
@tasxatzial Later on, we can consider adding specialized update functions for updating the `path` (which determines the rendered testing string argument) and/pr `expected.error` value, which are likely quite common transformations. But let's hold off on that until we've done more templates.

tasxatzial (approved):
Makes a lot more sense now

PR #781, opened 2025-01-18
sublist: sync tests

ErikSchierboom:


ErikSchierboom:
Sure. The purpose of `safe-str` is to prevent the templating system for formatting a specific value. In this particular case, the expected value is originally a `string`. If we were to do nothing, it would be rendered as "<expected>", with surrounding quotes. However, we don't want that here, as we just want a specific value to be rendered but without the quotes. With `safe-str`, the templating library will just render the value _as is_.

PR #782, opened 2025-01-18
Add `knapsack` exercise

ErikSchierboom:


tasxatzial (approved):
Nice. This one still has a `~` in the .tpl file.

Would it be easy to have the maps listed one below the other? I feel that would improve the presentation a bit.

tasxatzial:
Just ignore my previous suggestion. In order to make this happen i believe the vectors would need to be formatted based on their contents? Sounds like too much work.

ErikSchierboom:
I've updated. Let me know what you think.

tasxatzial:
Looks good! Somehow the food-chain files have made their appearance here.

PR #785, opened 2025-01-18
Add `food-chain` exercise

ErikSchierboom:


tasxatzial (commented):
Just by looking at your solution, i would probably bump the difficulty to 5.

Prerequisites: strings, conditionals, and either vectors or lists.

ErikSchierboom:
👍  I've updated

tasxatzial (approved):
Nice. I'm not sure if you intent to remove the `~` from all .tpl files. This one still has it at the end.

ErikSchierboom:
> I'm not sure if you intent to remove the ~ from all .tpl files. This one still has it at the end.

Removed! Thanks for noticing.

PR #786, opened 2025-01-18
space-age: sync tests

ErikSchierboom:


exercises/practice/space-age/test/space_age_test.clj
```diff
@@ -1,46 +1,40 @@
...
   [expected actual]
   (is (= (Math/round (* 100.0 expected))
          (Math/round (* 100.0 actual)))))
 
-(deftest age-in-earth-years
-  (rounds-to 31.69 (space-age/on-earth 1000000000)))
-
-(deftest age-in-mercury-years
-  (let [seconds 2134835688]
-    (rounds-to 67.65 (space-age/on-earth seconds))
```
ErikSchierboom, line 15: These earth-based parts of the data are not part of the canonical data, and do not belong here I feel

tasxatzial: Not only that, but the tests are incorrectly implemented. There should be only one function. That's why when I synced them last time, I left the test that throws an error out.

tasxatzial (commented):
This probably needs an update in the .tpl file (the last ~) ?

Also the space_age.clj file still has the old style function templates.

The functions all start with `on-` whereas in the tests they start with `on_`

ErikSchierboom:
All updated

tasxatzial (approved):
Nice!

PR #787, opened 2025-01-18
Add `killer-sudoku-helper` exercise

ErikSchierboom:


tasxatzial (commented):
Just by looking at the description and your solution, i would bump the difficulty to 5 or 6.

Prerequisites: conditionals, vectors or lists (or both), maybe numbers.

Function parameters: All 3 parameters are `cage` parameters. The general guideline in Clojure is to provide them as a map. So the input would look like `{:sum 5 :size 1 :exclude []}`. This will also allow practicing map destructuring.

ErikSchierboom:
Updated!

tasxatzial (approved):
To keep it simple for future exercises, please disregard what i said here about maps as general guideline. Let's just say we picked maps because it matches the canonical-data and it was a good opportunity for practicing maps. Another good option here would be to use a multi-arity function, with the exclude parameter being optional.

ErikSchierboom:
Let's keep multi arity functions in mind to find an opportunity to use them!

PR #790, opened 2025-01-19
Auto format rendered templates

ErikSchierboom:
I've been using [cljfmt](https://github.com/weavejester/cljfmt) lately and it has been really helpful getting my code formatted nicely. This PR is an experiment to see what happens if we use that library to auto format our generated templates.

The benefits are:

- Consistent style amongst exercises
- Templates become less of a hassle when trying to fix whitespace and do alignment (to me, these have been _the_ most annoying things when building templates)

The main downside is that you would have to adhere to the style used by cljfmt. We _could_ add support to opt-out of auto-formatting, but it would have to be the exception I feel.

### DIFF
--- a/docs/GENERATORS.md
+++ b/docs/GENERATORS.md
@@ -11,7 +11,8 @@ To generate a practice exercise's tests, the test generator:
 2. Uses `tests.toml` file to omit and excluded test cases
 3. Transforms the test cases (optional)
 4. Renders the test cases using the exercise's generator template
-5. Writes the rendered template to the exercise's test file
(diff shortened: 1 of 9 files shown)


ErikSchierboom:
This PR has two commits:

- The actual changes to support auto formatting
- What the changes to the rendered tests look like if they were auto-formatted

tasxatzial (approved):
Well, this looks perfectly fine to me 👍

ErikSchierboom:
@tasxatzial I've also included an update to the `GENERATORS.md` file.

PR #795, opened 2025-01-21
Add `game-of-life` exercise

ErikSchierboom:


exercises/practice/game-of-life/src/game_of_life.clj
```diff
@@ -0,0 +1,7 @@
+(ns game-of-life)
+
+(defn tick
+  "Return the next generation of the given matrix."
+  [matrix]
```
tasxatzial, line 5: Given that the exercise talks about cells, it's probably best if we adopt it for the docstring.
```suggestion
  "Returns the next generation of the cells."
  [cells]
```

tasxatzial (commented):
Splitting the tests into multiple lines seems kind of overkill to me. The matrices aren't really aligned so we may as well do something like

```
(is (= [[0 0 0]
        [0 0 0]
        [0 0 0]]
       (game-of-life/tick
         [[0 0 0]
          [0 1 0]
          [0 1 0]])))
```

One thing that bothers me a bit is that the canonical-data does not use standardized descriptions. It talks about matrices in two of the tests, and about cells in the rest of them. It should be talking about cells in all descriptions because a track might not even be using matrices for the cell representation.

ErikSchierboom:
Updated.

tasxatzial (approved):

PR #802, opened 2025-01-24
Update leiningen config to Clojure 1.12.0

tasxatzial:
This should allow to use methods as values and write things like `(map Integer/parseInt ...)` instead of `(map #(Integer/parseInt %) ...)`

See https://clojure.org/news/2024/09/05/clojure-1-12-0#method_values


ErikSchierboom (approved):
Nice

PR #803, opened 2025-01-29
Add `connect` exercise

tasxatzial:


tasxatzial:
Thanks! What do you think about the difficulty? Initially, I was considering a 7. I'm using a DFS algorithm, though it may not be necessary here. My solution is also quite verbose — there's no 'hacky' implementation involved. So, I'm unsure how simple this can actually get.

tasxatzial:
I think i'll go with hard. Checked solutions in other tracks and definitely looks like a lot of work. Also, plenty of medium exercises already.

ErikSchierboom:
I agree with that assessment

PR #810, opened 2025-02-18
binary-search: Update tests

tasxatzial:
* Changed to return -1 instead of some kind of error when value is not in the vector
* Updated function template

### DIFF
--- a/exercises/practice/binary-search/.docs/instructions.append.md
+++ b/exercises/practice/binary-search/.docs/instructions.append.md
@@ -1,0 +1,10 @@
+# Instructions append
+
+## Appendix
+
+~~~~exercism/note
+The instructions above are synchronized with a shared repository to ensure consistency across all language tracks.
+This appendix provides additional clarification or modifies the instructions as needed to better align with the goals of the Clojure track.
+~~~~
+
+For this exercise in the Clojure track, you may **assume the input is a vector**, as indicated by the tests.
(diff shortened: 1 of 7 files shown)


tasxatzial:
I'm trying to change the description, but i'm not having any success. I'm trying
```
(defn update-test-case [test-case]
  (assoc test-case :description "new description"))
```

ErikSchierboom (approved):

tasxatzial:
@ErikSchierboom Any idea why updating the description isn't working? Should the above code work, or am I missing something?

ErikSchierboom:
@tasxatzial That's because the description is overwritten later in the call chain:

```
(defn- test-case->data [idx node]
  (-> node
      (assoc :idx (inc idx)
             :description (str/join " ▶ " (:path node))
             :error (get-in node [:expected :error]))
      (dissoc :reimplements :comments :scenarios)))
```

That might not be the best design. Would you like me to have this happen _before_ the `update-test-case` function is called?

tasxatzial:
I was just looking at that particular piece of code.

> Would you like me to have this happen _before_ the `update-test-case` function is called?

Yes, please :)

ErikSchierboom:
It'll be next week before I have time

ErikSchierboom:
PR opened: https://github.com/exercism/clojure/pull/823

ErikSchierboom (approved):

PR #816, opened 2025-02-24
generator: rename template variable

ErikSchierboom:
This is done in preparation of changing the order of the generator calls, as we want to keep the `:description` from the canonical data and not overwrite it. I'm open to naming suggestions.

### DIFF
--- a/exercises/practice/acronym/.meta/generator.tpl
+++ b/exercises/practice/acronym/.meta/generator.tpl
@@ -4,6 +4,6 @@
 
 {{#test_cases.abbreviate}}
 (deftest acronym_test_{{idx}}
-  (testing {{description}}
+  (testing {{context}}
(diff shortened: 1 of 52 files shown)


tasxatzial:
So, what's the plan here? Could you give a brief explanation so I have some context? :)

ErikSchierboom:
Well, previously we overwrote the `description` property with the value used in the templates that contains the test case's description. As suggested previously, the description property should be set before the callbacks in the `generator.clj` are called, as that will allow you to update the description in those callbacks. But as some exercises use the `description`, we can't overwrite its value before that is called.

tasxatzial (approved):
Thanks. I think 'context' is fine, not going to be too picky about naming at this point. If anything comes up, we can update later.

PR #817, opened 2025-02-24
docs: add skip formatting doc

ErikSchierboom:


generators/src/templates.clj
```diff
@@ -53,18 +53,22 @@
     (mapv update-test-case-fn test-cases)
     test-cases))
 
-(defn- transform [slug test-cases]
+(defn- load-generator-ns [slug]
   (let [transform-file (paths/generator-clojure-file slug)]
-    (if (.exists transform-file)
+    (when (.exists transform-file)
```
tasxatzial, line 58: This returns `nil` if the transform file does not exist, and functions that have `generator-ns` as a parameter will most likely crash. Those functions don't seem to check if the transform file exists.

tasxatzial (commented):
Tried on the sieve PR, but couldn't get it to work at all. Formatting still happened. Not sure if i'm missing something.

Also, what's the expected result for an input like `[2,3,4]`?

tasxatzial:
> Tried on the sieve PR, but couldn't get it to work at all. Formatting still happened. Not sure if i'm missing something.

To clarify: Formatting still happened for the vectors. But L4 and L5 were both empty, so it looks that skipping formatting did work in some places.

ErikSchierboom:
> What if we allow auto formatting to always happen, and after it takes place, we do custom formatting in specific tests?

That's fine by me! Not sure how you'd do the custom formatting, but maybe it's so exceptional that that won't be a problem? Implementation-wise: another extension function in `generator.clj` I assume?

tasxatzial:
> Implementation-wise: another extension function in `generator.clj` I assume?

Probably, but i have absolutely no idea how we could go about it right now. I think we can keep auto formatting for now, and if it ever becomes a serious issue, we can reconsider continuing this PR or a new one. What do you think?

Right now i'm more concerned about the fact that certain exercises don't work at all with the templates. When you have some time, please try `etl`. Also, this is an exercise where custom formatting can be useful.

ErikSchierboom:
> Probably, but i have absolutely no idea how we could go about it right now. I think we can keep auto formatting for now, and if it ever becomes a serious issue, we can reconsider continuing this PR or a new one. What do you think?

Let's close this for now.

PR #823, opened 2025-02-25
Change order of operations

ErikSchierboom:


generators/src/templates.clj
```diff
@@ -64,9 +66,10 @@
       test-cases)))
 
 (defn- test-cases->data [slug test-cases]
-  (let [transformed (transform slug test-cases)
+  (let [augmented (map test-case->data test-cases)
+        transformed (transform slug augmented)
         grouped (group-by :property transformed)
-        data (update-vals grouped #(map-indexed test-case->data %))]
+        data (update-vals grouped #(map-indexed add-idx %))]
```
ErikSchierboom, line 72: This now applies the `test-case->data` function first, before the callbacks are being called.

ErikSchierboom:
@tasxatzial I've updated the `nth-prime` generator.

tasxatzial:
okay, so as far as I can tell, this seems to work as expected. Can you fill me in on what else has changed? I've noticed that the `:error` now seems to be updated using `(assoc test-case :error "error-message")`, whereas before it was `(assoc-in test-case [:expected :error] "error-message")`

ErikSchierboom:
> okay, so as far as I can tell, this seems to work as expected. Can you fill me in on what else has changed? I've noticed that the :error now seems to be updated using (assoc test-case :error "error-message"), whereas before it was (assoc-in test-case [:expected :error] "error-message")

That is because I've changed the logic that executed before the function extensions are called to this:

```
(defn- test-case->data [node]
  (-> node
      (assoc :context (str/join " ▶ " (:path node))
             :error (get-in node [:expected :error]))
      (dissoc :reimplements :comments :scenarios)))
```

This used to be done _after_ the extension point, but now it run before it. Now that I think of it, maybe it would be better to have the `error` bit _after_ the extension point. And maybe the `scenarios` field should be kept and not discarded?

ErikSchierboom:
@tasxatzial I've done some refactoring and I feel like this new strategy is a lot better.

tasxatzial:
This is what i have so far. This replaces "list" with "vector" everywhere in the path.

```
(ns list-ops-generator
  (:require [clojure.string :as str]))

(defn update-path [path]
  (->> path
       (map #(str/replace % #"list" "vector"))
       (str/join " ▶ ")))

(defn update-test-case [test-case]
  (->> (:path test-case)
       update-path
       (assoc test-case :context)))                                    
```

It does work, but i'm not sure how you feel about the manual join.

ErikSchierboom:
> It does work, but i'm not sure how you feel about the manual join.

I'm fine with that to be honest. It should be an exceptional case, so I mind it.

exercises/practice/grains/.meta/generator.clj
```diff
@@ -8,4 +8,4 @@
 (defn update-test-case [test-case]
   (-> test-case
       (update :expected update-expected)
-      (update :path #(take-last 1 %))))
+      (update :context (:description test-case))))
```
tasxatzial, line 11:
```suggestion
      (assoc :context (:description test-case))))
```

tasxatzial (changes requested):

tasxatzial (approved):
I believe this is good to go. Hopefully I haven't missed anything.

PR #829, opened 2025-03-06
flatten-array: Update to latest

tasxatzial:
Should be merged after #823


tasxatzial:
I've added an append file to let people that they are free to assume more than vectors (just as the flatten function does in clojure.core)

ErikSchierboom (approved):

PR #840, opened 2025-03-22
allergies: add generators and regenerate tests

tasxatzial:
[no important files changed]

While `(safe-str (keyword "eggs"))` results in `:eggs` in final test file, when mapping over arrays the result looks like this

```
[
#object[com.github.jknack.handlebars.Handlebars$SafeString 0x135f160e ":eggs"]
#object[com.github.jknack.handlebars.Handlebars$SafeString 0x230e163b ":shellfish"]
]
```

Not sure how i could make this work, i switched to a different approach with `symbol` and without `safe-str`:

`#(symbol (str (keyword %)))`

does seem to do the trick.

@ErikSchierboom Thoughts or suggestions?

### DIFF
--- a/exercises/practice/allergies/.meta/generator.clj
+++ b/exercises/practice/allergies/.meta/generator.clj
@@ -1,0 +1,25 @@
+(ns allergies-generator
+  (:require [clojure.string :as str]))
+
+(defn- create-context [test-case]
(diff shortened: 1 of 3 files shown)


ErikSchierboom (approved):
I think this is still very readable

